PR #61, opened 2025-12-30
Add double-tap gesture for play/pause on Apple Watch

jamesm131:
## Summary
- Adds `.handGestureShortcut(.primaryAction)` to the play/pause button in the Watch app player
- Enables the watchOS 10+ double-tap gesture to toggle playback
- Uses `#available(watchOS 11.0, *)` check for backward compatibility with watchOS 10

## Test plan
- [x] Build and deploy to Apple Watch running watchOS 11+
- [x] Open the player with an audiobook
- [x] Perform the double-tap gesture (pinch thumb and index finger together twice)
- [x] Verify playback toggles between play and pause


jeremygrenier:
@jamesm131 looks good to me. Thank you for adding it.

PR #130, opened 2026-01-28
Update shake sensitivity labels for clarity

DeeKahy:
Renamed shake sensitivity options from technical sensitivity levels to intuitive gesture descriptions. The previous labels (Off, Very High, High, Medium, Low, Very Low) were confusing because "High" sensitivity required a gentle or low shaking strength while "Low" required a vigorous/a lot of shaking.

#### Files Changed
- `PlayerPreferencesView.swift`
- `UserPreferences.swift` - I hope that was correct, I am not sure what it is used for.

#### Alternative Options
If you don't like these labels, here are two alternatives:

Off, Barely a Tap, Light Shake, Normal Shake, Hard Shake, Very Hard Shake

Off, Ultra Sensitive, Very Sensitive, Sensitive, Less Sensitive, Least Sensitive

### DIFF
--- a/AudioBooth/AudioBooth/Screens/Settings/Preferences/Player/PlayerPreferencesView.swift
+++ b/AudioBooth/AudioBooth/Screens/Settings/Preferences/Player/PlayerPreferencesView.swift
@@ -161,7 +161,7 @@ struct PlayerPreferencesView: View {
         }
         .font(.caption)
 
-        Picker("Shake to Reset", selection: $preferences.shakeSensitivity) {
+        Picker("Shake Sensitivity to Reset", selection: $preferences.shakeSensitivity) {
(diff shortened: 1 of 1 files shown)


jeremygrenier:
@DeeKahy Thank you for the PR!

I like the direction you're taking. What do you think of renaming the feature itself to **Shake Sensitivity to Reset** and keeping the original sensitivity levels like **Very Low, Low, Medium, High, Very High**? That way the context makes the sensitivity values clearer.

Until we have localization in the app, I'd like to keep the wording simple for non-native English speakers.

DeeKahy:
That definitely also works, and I do agree the word was probably a little hard. I have pushed the change to this pull request.

 Also on that note, I can most likely help with the Danish and mostly also German translations when the time comes.

PR #131, opened 2026-02-02
Translation: Danish

DeeKahy:
Progress on the Danish translation.

Obviously, translation can be subjective, and some of my translations are definitely a little dicey because the there is no direct equivalent.


DeeKahy:
Sorry for all the commit messages. Probably a good idea to squash on merge.

I do not know why xcode decided to format strings i did not touch, sorry about that.

jeremygrenier:
I rebased and merged your translation. Thank you for doing it!

PR #157, opened 2026-02-26
Fix watch progress being overwritten by stale iPhone data

marcosgarrido:
## What was happening

When listening from the Apple Watch, progress updates were sent to the iPhone,  
but the iPhone did not immediately update its local `MediaProgress`.

Because of that:

- The iPhone could send outdated progress back to the watch.
- In some cases, opening a book on the watch would show progress at `0`
  or at an older position.
- If the same book screen was already open on the watch,
  progress updates coming from the iPhone were not reflected
  in the already displayed player UI.

---

## What this PR changes

- Includes `bookID` in the `reportProgress` message.
- Updates local `MediaProgress` on iOS immediately when progress is received from the watch.
- Refreshes the watch context after syncing to ensure both sides stay aligned.
- Adds a reactive progress observer in the Watch player so that
  progress updates are reflected even if the user stays on the same book.

---

## Result

Progress now remains consistent across Apple Watch, iPhone, and server.

Books no longer reset to `0`, revert to stale positions,
or require switching books to reflect updated progress.

### DIFF
--- a/AudioBooth/AudioBooth Watch App/Screens/Player/BookPlayerModel.swift
+++ b/AudioBooth/AudioBooth Watch App/Screens/Player/BookPlayerModel.swift
@@ -646,10 +646,11 @@ final class BookPlayerModel: PlayerView.Model {
   }
 
   private func syncToLatestKnownProgressIfNeeded() {
-    guard !isPlaying else { return }
-
-    let latest = connectivityManager.progress[bookID] ?? current
-    guard abs(latest - current) > 0.5 else { return }
+    guard
+      !isPlaying,
+      let latest = connectivityManager.progress[bookID],
+      latest - current > 0.5
+    else { return }
 
     current = latest
 
--- a/AudioBooth/AudioBooth/Services/WatchConnectivityManager.swift
+++ b/AudioBooth/AudioBooth/Services/WatchConnectivityManager.swift
@@ -447,10 +447,6 @@ extension WatchConnectivityManager: WCSessionDelegate {
           currentTime: currentTime
         )
 
-        await MainActor.run {
-          self.refreshProgress()
-        }
-
         AppLogger.watchConnectivity.debug("Synced watch progress: \(currentTime)s")
       } catch {
         AppLogger.watchConnectivity.error("Failed to sync watch progress: \(error)")


AudioBooth/AudioBooth/Services/WatchConnectivityManager.swift
```diff
@@ -420,19 +422,35 @@ extension WatchConnectivityManager: WCSessionDelegate {
...
-  private func handleProgressReport(sessionID: String, currentTime: Double, timeListened: Double) {
+  private func handleProgressReport(bookID: String, sessionID: String, currentTime: Double, timeListened: Double) {
     Task {
       do {
+        let duration =
+          (try? LocalBook.fetch(bookID: bookID))?.duration
+          ?? (try? MediaProgress.fetch(bookID: bookID))?.duration
+          ?? 0
+
+        let safeDuration = (duration > 0) ? duration : max(currentTime, 1)
```
jeremygrenier: I'm not a huge fan of fetching `LocalBook` or `MediaProgress` here. I think we can have `reportProgress` send the `duration` and the `progress` instead of recomputing it on the app side.

AudioBooth/AudioBooth Watch App/WatchConnectivityManager.swift
```diff
@@ -272,7 +274,7 @@ final class WatchConnectivityManager: NSObject, ObservableObject {
           duration: duration,
           chapters: chapters,
           tracks: tracks,
-          currentTime: 0
+          currentTime: currentTime
```
jeremygrenier:
```suggestion
          currentTime: self.progress[id, default: 0]
```

AudioBooth/AudioBooth Watch App/WatchConnectivityManager.swift
```diff
@@ -263,6 +264,7 @@ final class WatchConnectivityManager: NSObject, ObservableObject {
         let coverURL = (response["coverURL"] as? String).flatMap { URL(string: $0) }
         let sessionID = response["sessionID"] as? String
 
+        let currentTime = self.progress[id] ?? 0
```
jeremygrenier: 🪓

AudioBooth/AudioBooth Watch App/Screens/Player/BookPlayerModel.swift
```diff
@@ -637,6 +641,44 @@ final class BookPlayerModel: PlayerView.Model {
...
+      }
+      .store(in: &cancellables)
+
+    syncToLatestKnownProgressIfNeeded()
+  }
+
+  private func syncToLatestKnownProgressIfNeeded() {
+    guard !isPlaying else { return }
+
+    let latest = connectivityManager.progress[bookID] ?? current
```
jeremygrenier:
```suggestion
    guard 
      !isPlaying, 
      let latest = connectivityManager.progress[bookID],
      latest - current > 0.5,
    else { return }
```

I think we progress do not exist we can skip the sync no? Also if latest < at current we should probably keep current. So no need for the abs.

AudioBooth/AudioBooth Watch App/Screens/Player/BookPlayerModel.swift
```diff
@@ -637,6 +641,44 @@ final class BookPlayerModel: PlayerView.Model {
...
+    guard let track = findTrack(for: latest, in: tracks) else { return }
+
+    if track.index != currentTrackIndex {
+      loadTrack(track, seekTo: latest)
+    } else {
+      let trackStartTime = calculateTrackStartTime(trackIndex: track.index)
+      let seekTime = max(0, latest - trackStartTime)
+      player.seek(to: CMTime(seconds: seekTime, preferredTimescale: 1000))
+    }
+  }
```
jeremygrenier, line 680: Not a blocker for this PR but I think we improve and refactor our seek logic between this code, `findTrack`, `loadTrack` and `seekToChapter`.

AudioBooth/AudioBooth/Services/WatchConnectivityManager.swift
marcosgarrido: Updated as suggested. Duration is now sent from the watch instead of being fetched on the iPhone side.

AudioBooth/AudioBooth Watch App/WatchConnectivityManager.swift
marcosgarrido: Applied the suggested change.

AudioBooth/AudioBooth Watch App/WatchConnectivityManager.swift
marcosgarrido: Applied the suggested change.

AudioBooth/AudioBooth Watch App/Screens/Player/BookPlayerModel.swift
marcosgarrido: Applied the suggested change.

AudioBooth/AudioBooth Watch App/Screens/Player/BookPlayerModel.swift
marcosgarrido: Extracted the shared seek logic to reduce duplication between syncToLatestKnownProgressIfNeeded and seekToChapter. We can further refactor findTrack/loadTrack in a follow-up PR if needed.

AudioBooth/AudioBooth Watch App/Screens/Player/BookPlayerModel.swift
jeremygrenier: I do not see any changes. Did you push it?

AudioBooth/AudioBooth/Services/WatchConnectivityManager.swift
```diff
@@ -420,19 +424,36 @@ extension WatchConnectivityManager: WCSessionDelegate {
...
         try await Audiobookshelf.shared.sessions.sync(
           sessionID,
           timeListened: timeListened,
           currentTime: currentTime
         )
+
+        await MainActor.run {
+          self.refreshProgress()
+        }
+
```
jeremygrenier: Any reason to call refresh progress here? The watch sent the latest progress so we probably do not need to send it back.

AudioBooth/AudioBooth Watch App/Screens/Player/BookPlayerModel.swift
marcosgarrido: Sorry! I accidentally reverted this while making another change. I’ve fixed it now.

AudioBooth/AudioBooth/Services/WatchConnectivityManager.swift
marcosgarrido: True! removed :)

jeremygrenier:
Looks good to me. Thank you for the contribution, and feel free to adjust any of the watchOS code. It’s quite painful to work with and test.

PR #160, opened 2026-02-28
Fix Now Playing metadata sync in CarPlay

marcosgarrido:
## Fix: CarPlay Now Playing UI Desync

This PR fixes an issue where the CarPlay interface (titles, progress, and duration) fails to update correctly when switching chapters.

### 🛠 Changes
- **Dynamic Identifiers:** Updated `ExternalContentIdentifier` to include chapter-specific IDs, forcing CarPlay to recognize track transitions.

### ✅ Result
- The CarPlay UI now updates titles and progress bars instantly upon chapter changes.

### DIFF
--- a/AudioBooth/AudioBooth/Services/NowPlayingManager.swift
+++ b/AudioBooth/AudioBooth/Services/NowPlayingManager.swift
@@ -166,11 +166,12 @@ final class NowPlayingManager {
   func update() {
     guard let player, let mediaProgress else { return }
 
-    var info = [String: Any]()
-
     info[MPMediaItemPropertyArtwork] = artwork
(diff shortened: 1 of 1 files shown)


jeremygrenier:
Oh I missed

> Dynamic Identifiers: Updated ExternalContentIdentifier to include chapter-specific IDs, forcing CarPlay to recognize track transitions.

but everything else should have no impact. I'm good trying to adjust `ExternalContentIdentifier` if that seems to be fixing the refreshing issue.

marcosgarrido:
I’ve been testing this further and I’ve verified that the UI refresh issue can be solved in two different ways.

I've confirmed that using a new local dictionary fixes the problem, but using dynamic IDs also resolves it on its own. Since either change independently solves the issue, there’s no need to apply both at the same time.

I’ll stick with the dynamic IDs as it’s the cleanest approach to ensure the UI stays in sync during chapter changes.

Issue #166, opened 2026-03-04
[Feature Request] Per-Book Narration Speed


octopotato:
Add the ability to set a default narration speed in Settings, and allow each book to remember its own playback speed independently. Changing the speed on one book should not affect the speed of another — identical to how Audible recently implemented per-book speed.

**Current behavior:**
A single global playback speed applies to all books. Changing speed while listening to book A also changes it for book B.

**Desired behavior:**
- A "Default Speed" option in Settings > Player that sets the initial speed for new books (e.g. 1.0×)
- Each book remembers its own speed across sessions
- Opening a book for the first time uses the default speed
- Changing speed on book A has no effect on book B's speed
- Episodes/podcasts continue using the global default

PR #167, opened 2026-03-04
Per-Book Narration Speed

octopotato:
- closes https://github.com/AudioBooth/AudioBooth/issues/166

Per-book narration speed, Audible-style. Each book remembers its own playback speed independently, so changing speed on one book doesn't affect another. Books without a saved speed fall back to a configurable global default in Settings > Player.

I added `playbackSpeed: Double?` to `MediaProgress` so speed is stored alongside all the other progress data in SwiftData. `SpeedPickerSheetViewModel` reads/writes it directly.

I also added `playbackRate` to the API model (`User.MediaProgress`) so if/when ABS supports it server-side, per-book speed will sync across devices automatically. I opened [advplyr/audiobookshelf#5106](https://github.com/advplyr/audiobookshelf/pull/5106) for that, but nothing here depends on it. Sending speed back to the server is a small follow-up once that lands.

Settings gets a new "Narration Speed" section with a "Default" row. Existing users' global speed migrates to `defaultPlaybackSpeed` on first launch so nothing resets to 1.0×.

(purple boxes are my additions and not in app)

| Settings | Book |
|--------|--------|
| image | Screenshot 2026-03-03 at 10 52 24 PM |

> [!IMPORTANT]
> I generated the demo with the old UI for the global settings being the floating picker sheet, I don't want to make another demo :)
 
https://github.com/user-attachments/assets/a5de0379-2f7e-492c-b8e0-04940fcf1b9f

> [!NOTE]
> I don't have any podcasts so I wasn't able to test that, but per-episode speed should work the same way since each episode has its own `MediaProgress` record.

### DIFF
--- a/AudioBooth/AudioBooth/Screens/BookPlayer/Sheets/SpeedPickerSheetModel.swift
+++ b/AudioBooth/AudioBooth/Screens/BookPlayer/Sheets/SpeedPickerSheetModel.swift
@@ -1,13 +1,10 @@
 import AVFoundation
-import Combine
 import Models
 import SwiftUI
 
 final class SpeedPickerSheetViewModel: FloatPickerSheet.Model {
   private let sharedDefaults = UserDefaults(suiteName: "group.me.jgrenier.audioBS")
   private let mediaProgress: MediaProgress?
-  private let preferences = UserPreferences.shared
-  private var cancellables = Set<AnyCancellable>()
 
   let player: AVPlayer
 
@@ -34,21 +31,6 @@ final class SpeedPickerSheetViewModel: FloatPickerSheet.Model {
       presets: [0.7, 1.0, 1.2, 1.5, 1.7, 2.0],
       defaultValue: 1.0
     )
-
-    observeDefaultSpeedChanges()
-  }
-
-  private func observeDefaultSpeedChanges() {
-    preferences.objectWillChange
-      .debounce(for: .milliseconds(100), scheduler: RunLoop.main)
-      .sink { [weak self] _ in
-        guard let self, mediaProgress?.playbackSpeed == nil else { return }
-
-        let newDefault = preferences.defaultPlaybackSpeed
-        guard newDefault > 0 else { return }
-        applySpeed(newDefault)
-      }
-      .store(in: &cancellables)
   }
 
   override func onIncrease() {
@@ -62,14 +44,11 @@ final class SpeedPickerSheetViewModel: FloatPickerSheet.Model {
   }
 
   override func onValueChanged(_ newValue: Double) {
-    mediaProgress?.playbackSpeed = (newValue / 0.05).rounded() * 0.05
-    applySpeed(newValue)
-  }
-
-  private func applySpeed(_ newValue: Double) {
     let rounded = (newValue / 0.05).rounded() * 0.05
     value = rounded
     let floatValue = Float(rounded)
+
+    mediaProgress?.playbackSpeed = rounded
     sharedDefaults?.set(floatValue, forKey: "playbackSpeed")
 
     player.defaultRate = floatValue
--- a/AudioBooth/AudioBooth/Screens/BookPlayer/BookPlayerModel.swift
+++ b/AudioBooth/AudioBooth/Screens/BookPlayer/BookPlayerModel.swift
@@ -1203,6 +1203,7 @@ extension BookPlayerModel {
 
         mediaProgress.lastPlayedAt = Date()
         mediaProgress.lastUpdate = Date()
+        mediaProgress.playbackSpeed = speed.value
         if mediaProgress.duration > 0 {
           mediaProgress.progress = mediaProgress.currentTime / mediaProgress.duration
         }


jeremygrenier:
Thank you for the PR.

I think the per-book playback speed should probably be part of `LocalBook` / `LocalEpisode` or `MediaProgress`. In an ideal world, the ABS server would manage it and associate it with `MediaProgress` so you can keep your playback speed across apps.

Regarding the player preferences, I think we can use a stepper like `EbookReaderPreferences` does. I don’t see much value in showing the full `FloatPickerSheet` when the goal is just to set a default value.

octopotato:
> I think the per-book playback speed should probably be part of `LocalBook` / `LocalEpisode` or `MediaProgress`. In an ideal world, the ABS server would manage it and associate it with `MediaProgress` so you can keep your playback speed across apps.

Love this idea, didn't even think of that! The ABS server has an open request for this that's been unaddressed since 2022
- https://github.com/advplyr/audiobookshelf/issues/1173

Let me see if I can cobble together a PR on ABS, it _seems_ like a small amount of work 🤞🏽.

---

> Regarding the player preferences, I think we can use a stepper like `EbookReaderPreferences` does. I don’t see much value in showing the full `FloatPickerSheet` when the goal is just to set a default value.

Oo good thought, I mainly went with the full `FloatPickerSheet` because that's how audible does it too, I'll play around with it!

octopotato:
- https://github.com/advplyr/audiobookshelf/pull/5106

Hopefully it gets accepted 🤞🏽

jeremygrenier:
@octopotato In the meantime you can still add it to `MediaProgress` model on the iOS side and use it for the app. It will be pretty trivial to adjust the logic if your ABS server PR get approved.

AudioBooth/AudioBooth/Screens/BookPlayer/Sheets/SpeedPickerSheetModel.swift
jeremygrenier, line 1: I'm a bit confused by all the changes on this file.

I think ideally it should stay the same except for the init part.

What I think we should do is update `BookPlayerModel.updateMediaProgress` to make sure we also update the playback progress at the same time so when ABS is ready to support it we will just have to send it to the BE at the same time.

octopotato: I moved the default speed observer and persistence logic to `BookPlayerModel` so the speed picker is mostly unchanged.

The main driver for the extra changes in the speed picker is that I wanted the currently playing book to update when you change the default speed in Settings, but only if the book doesn't have a per-book speed set. So `onValueChanged` writes `mediaProgress.playbackSpeed` (marks it as explicit), and applySpeed is the same thing minus that write, used by the observer so the default doesn't get baked in as a per-book speed.

I looked at how `PlaybackProgressViewModel` observes preference changes for `showFullBookDuration` and followed the same pattern (debounced `objectWillChange` sink).

If the live default propagation isn't worth the complexity, I can drop it and the speed picker changes shrink to just the init, matching `VolumeLevelSheetViewModel`'s pattern. The default would just apply the next time you open a book. It's a pretty niche use case and I'm not sure how noticeable it would be on normal app use. But during testing it was something I noticed immediately and was kinda annoying that I had to switch books before the new "default" speed got applied.

Let me know what you think!

jeremygrenier: Oh sorry for the confusion I missed that. Make sense to have the logic in `SpeedPickerSheetViewModel` then like you did first.

AudioBooth/AudioBooth/Services/UserPreferences/UserPreferences.swift
```diff
@@ -201,6 +205,14 @@ final class UserPreferences: ObservableObject {
     default: break
     }
   }
+
+  private func migrateGlobalPlaybackSpeed() {
```
jeremygrenier: ❓ Any reason to rename `playbackSpeed` to `defaultPlaybackSpeed`? If we keep `playbackSpeed` then we can drop this migration and keep things simpe no?

Models/Sources/Models/MediaProgress.swift
```diff
@@ -307,6 +312,10 @@ extension MediaProgress {
           local.lastUpdate = remote.lastUpdate
         }
 
+        if let remoteSpeed = remote.playbackSpeed {
+          local.playbackSpeed = remoteSpeed
+        }
+
```
jeremygrenier: I think we would want to update the `playbackSpeed` only if `remote.lastUpdate > local.lastUpdate` no?

jeremygrenier: We will also need to check if `BookPlayer` playback speed has changed when the user tap on play. To be honest I'm not sure we want to update `‎API/Sources/API/Models/User.swift` and this code part of this PR.

It is possible ABS server PR will not get merged or they will push back on another solution. I would rather avoid having dead code and add after ABS merged it.

octopotato: yea I was being super optimistic and hoping it'll ship soon/quickly, but I'm starting to lose hope for the ABS change lol.. I'll remove the dependency/assumption on what the API might look

AudioBooth/AudioBooth/Services/UserPreferences/UserPreferences.swift
octopotato: good call!

AudioBooth/AudioBooth/Screens/Settings/Preferences/Player/PlayerPreferencesView.swift
```diff
@@ -330,6 +330,31 @@ struct PlayerPreferencesView: View {
     .listRowSeparator(.hidden)
     .listSectionSpacing(.custom(12))
 
+    Section {
```
octopotato, line 333: I kinda don't like that this isn't the full floating menu for the speed setting that the player (like what audible does) has because it's less convenient to go from 1x to 2x for the global default. But since this is a one time action it's _fine_ and just a one time inconvenience.. I'm probably feeling it more because of my testing

AudioBooth/AudioBooth/Screens/BookPlayer/Sheets/SpeedPickerSheetModel.swift
```diff
@@ -34,10 +62,14 @@ final class SpeedPickerSheetViewModel: FloatPickerSheet.Model {
   }
 
   override func onValueChanged(_ newValue: Double) {
+    mediaProgress?.playbackSpeed = (newValue / 0.05).rounded() * 0.05
+    applySpeed(newValue)
+  }
```
octopotato: we only want to set the per-book playback if it changed in the player and not in the global setting

octopotato: alternatively we could keep `onValueChanged` and add a `persist` flag or something to decide when to save `mediaProgress?.playbackSpeed` or not

jeremygrenier: > we only want to set the per-book playback if it changed in the player and not in the global setting

Any reason to do that? I feel we can always set `mediaProgress?.playbackSpeed` at the speed the user is currently listening regardless if it is coming from user preferences or not no?

octopotato: If we always write it then the global default basically becomes a one-shot thing, as soon as a book is played its speed gets locked in and changing the global default later won't affect it.

I think we want `nil` to mean "follow the global default" so books without an explicit per-book speed always pick up whatever the current default is. That's how Audible handles it, and it also means we won't accidentally push a per-book speed to ABS when the user only changed the global setting.

Basically we need a way to differentiate between "this is an action the user took on this specific book" vs "this is an action the user took for all books"

jeremygrenier: I can see some value in that, but people are likely going to update the default rate only once. If we start questioning when we should update it or not, it can raise a lot of questions.

- If I change the playback speed of a book, should it update the default playback speed if it is not already set?
- If I change the default playback speed, should it adjust the playback speed of the current book?

One advantage of always setting the value is that if I set a default speed to 2x in AudioBooth and then move to the ABS web player (once ABS Server supports playback speed), I will continue to enjoy the same speed I was listening to on my phone. If we do not do that, it means you would need to make sure all your clients use the same default speed.

For me, `playbackSpeed` should always be defined as the actual speed you are listening to the book at, regardless of whether it is the default or not.

octopotato: yea that's a good point and a key difference between Audible and ABS. Unless ABS _also_ allows setting the global default, you'd end up with different book speed rates as you switch apps/clients. I'll update it! It'll also simplify the logic a bit

PR #169, opened 2026-03-05
Fix/carplay offline not updating

marcosgarrido:
### Fix CarPlay offline library not updating

The CarPlay offline tab did not refresh when downloads complete or when offline books were removed locally.

This change follows the same observation pattern used in `OfflineListViewModel`, observing `LocalBook.observeAll()` to reload the list when the local library changes.

Additionally, selecting a book from the offline tab now starts playback automatically, matching the behavior implemented in `CarPlayLibrary`.

Side question: I noticed there’s a `try? await Task.sleep(for: .milliseconds(500))` after selecting an offline book. Is this delay intentional (e.g. to wait for the player to be ready), or should it be removed?

Thank you!

### DIFF
--- a/AudioBooth/AudioBooth/CarPlay/CarPlayController.swift
+++ b/AudioBooth/AudioBooth/CarPlay/CarPlayController.swift
@@ -1,7 +1,7 @@
 @preconcurrency import CarPlay
 import Foundation
 
-class CarPlayController: NSObject, CPInterfaceControllerDelegate {
+class CarPlayController {
   private let interfaceController: CPInterfaceController
 
   private let tabBar: CarPlayTabBar
@@ -12,17 +12,6 @@ class CarPlayController: NSObject, CPInterfaceControllerDelegate {
 
     nowPlaying = .init(interfaceController: interfaceController)
     tabBar = .init(interfaceController: interfaceController, nowPlaying: nowPlaying)
-    
-    super.init()
-
-    self.interfaceController.delegate = self
-    
     try await interfaceController.setRootTemplate(tabBar.template, animated: false)
-  }
-  
-  func templateWillAppear(_ template: CPTemplate, animated: Bool) {
-    Task {
-      await tabBar.handleWillAppear(for: template)
-    }
   }
 }
--- a/AudioBooth/AudioBooth/CarPlay/CarPlayOffline.swift
+++ b/AudioBooth/AudioBooth/CarPlay/CarPlayOffline.swift
@@ -4,38 +4,38 @@ import Foundation
 import Models
 import Nuke
 
-final class CarPlayOffline: CarPlayReloadable  {
+final class CarPlayOffline: CarPlayPageProtocol {
   private let interfaceController: CPInterfaceController
   private weak var nowPlaying: CarPlayNowPlaying?
   private var currentPlayerCancellable: AnyCancellable?
   private let downloadManager = DownloadManager.shared
-  
+
   let template: CPListTemplate
-  
+
   init(interfaceController: CPInterfaceController, nowPlaying: CarPlayNowPlaying) {
     self.interfaceController = interfaceController
     self.nowPlaying = nowPlaying
-    
+
     let title = String(localized: "Offline")
     template = CPListTemplate(title: title, sections: [])
     template.tabTitle = title
     template.tabImage = UIImage(systemName: "arrow.down.circle.fill")
-    
+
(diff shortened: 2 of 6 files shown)


jeremygrenier:
Thanks for the contribution! A few notes:

**On the observation approach:** We'd prefer to keep CarPlay classes free of ongoing observations for simplicity. Instead of observing `LocalBook.observeAll()`, we'd like to reload the books list when the template becomes visible using `templateWillAppear` or an equivalent lifecycle callback. This keeps the refresh logic simple and avoids managing a background `Task` and its cancellation.

**On the `Task.sleep` delay:** The 500ms delay is intentional. The now playing screen depends on the `AVAudioSession` being properly initialized after `setCurrent`, and navigating before that was causing issues. Please keep it as-is.

**On `PlayerManager.shared.play()`:** This addition should be reverted. CarPlay audio apps should not trigger playback automatically on selection; the system and the user control when playback starts. The existing behavior (without the explicit `play()` call) is consistent with how other CarPlay audio apps handle this.

marcosgarrido:
Thanks for the review!

I've updated the PR accordingly:

- Removed the `LocalBook.observeAll()` observation and now reload the Offline list from `templateWillAppear` when the template becomes visible.
- Removed the `PlayerManager.shared.$current` observation in `CarPlayOffline`
- Kept the `try? await Task.sleep(for: .milliseconds(500))` delay.
- Removed the explicit `PlayerManager.shared.play()` call (in both `CarPlayOffline` and `CarPlayLibrary`).

Let me know if anything else should be adjusted. :)

AudioBooth/AudioBooth/CarPlay/CarPlayController.swift
```diff
@@ -1,7 +1,7 @@
 @preconcurrency import CarPlay
 import Foundation
 
-class CarPlayController {
+class CarPlayController: NSObject, CPInterfaceControllerDelegate {
```
jeremygrenier: I don't think that is what we want. We want to reload when the offline tab appears, not when the CarPlay interface appears.

Probably by using [`tabBarTemplate(_:didSelect:)`](https://developer.apple.com/documentation/carplay/cptabbartemplatedelegate/tabbartemplate(_:didselect:)) so each time a tab is selected we can reload it to make sure it is up to date.

AudioBooth/AudioBooth/CarPlay/CarPlayLibrary.swift
```diff
@@ -66,7 +66,6 @@ final class CarPlayLibrary {
 
   private func onBookSelected(_ book: Book) {
     PlayerManager.shared.setCurrent(book)
-    PlayerManager.shared.play()
```
jeremygrenier, line 69: Sorry for my ask before maybe it is fine to keep the `PlayerManager.shared.play()` it seems to be what we are doing except for offline so we should align everywhere.

AudioBooth/AudioBooth/CarPlay/CarPlayOffline.swift
```diff
@@ -21,17 +20,15 @@ final class CarPlayOffline {
     template.tabTitle = title
     template.tabImage = UIImage(systemName: "arrow.down.circle.fill")
 
-    currentPlayerCancellable = PlayerManager.shared.$current.sink { [weak self] _ in
```
jeremygrenier, line 24: Do not remove `PlayerManager.shared.$current.sink` my ask was to not use observation for checking download because it is not something you should be doing when driving,

However this observation make sure to adjust the list when you start playing a book from CarPlay.

AudioBooth/AudioBooth/CarPlay/CarPlayController.swift
marcosgarrido: Got it. Thanks!

I ran some tests using `tabBarTemplate(_:didSelect:)` and noticed a small difference: the list won't reload when returning from NowPlaying, since didSelect is only triggered when the tab itself is tapped. If a book is downloaded while in NowPlaying, you have to go to the Home tab and then tap the Offline tab for the list to refresh.

Previously `templateWillAppear` also handled that case. Let me know if you'd still prefer to keep the `tabBarTemplate(_:didSelect:)` approach.

jeremygrenier: I think we can keep both and add a `onAppear` on `CarPlayOffline` when one or the other is called.

Issue #179, opened 2026-03-09
Book metadata not sending correctly via Bluetooth or USB to non-Carplay car display.


b-harting:
## Description
When playing a book via Bluetooth to my car, the book title is being sent as Artist, and nothing is being send as Title.

## Steps to Reproduce
1. Connect to your car stereo as via Bluetooth, not Carplay.
2. Play any book

## Expected Behavior
- Track title should be the current chapter
- Artist should be the author(s).
- Album should be the book title.

## Actual Behavior
The metadata appears as follows:
  - Track title will be the current chapter (this is correct)
  - Artist will be the title of the book (incorrect)
  - Album (Book) title will be empty.

## Screenshots
![Image](https://github.com/user-attachments/assets/2e7f6230-95c6-4f28-834e-fe2b35664180)

## Device Information
- **Platform:** iPhone
- **Device Model:** iPhone 14
- **OS Version:** iOS 26.1
- **AudioBooth Version:** Both 1.7 (current release) and 1.8 (1772951043) (current beta)
- **Audiobookshelf Server Version:** 2.32.1

## Additional Context
This data shows as expected via other apps like Audible and Prologue, so it's technically possible.

b-harting:
I tried connecting via USB today, and also had metadata issues.

![Image](https://github.com/user-attachments/assets/8b263f96-913f-4e84-be18-a151931083c1)

The cover is being sent, but no other info.

PR #181, opened 2026-03-11
Fix Now Playing metadata for Bluetooth/USB car displays

octopotato:
- #179

I found the bug in `NowPlayingManager.update()`. When showing per-chapter info, `MPMediaItemPropertyAlbumTitle` was never set, so car displays and Bluetooth AVRCP showed an empty album field. I added the author to `MPMediaItemPropertyAlbumTitle` in chapter mode so all three fields are populated for external displays.

When `showFullBookDuration` is on (full book mode), album title gets cleared so stale values don't carry over.

### Lock Screen (chapter duration)
image

### Lock Screen (book duration)
image

#### Before
| Book duration | Chapter duration |
|--------|-------|
| image | image |

#### After
| Book duration | Chapter duration |
|--------|-------|
| image | image |

I don't have a Bluetooth/USB car setup to test AVRCP directly, but it reads from the same `MPNowPlayingInfoCenter` dictionary so the data should be correct there too.

### DIFF
--- a/AudioBooth/AudioBooth/Services/NowPlayingManager.swift
+++ b/AudioBooth/AudioBooth/Services/NowPlayingManager.swift
@@ -175,8 +175,8 @@ final class NowPlayingManager {
 
     if !preferences.showFullBookDuration, let chapters, let current = chapters.current {
       info[MPMediaItemPropertyTitle] = current.title
-      info[MPMediaItemPropertyArtist] = author
-      info[MPMediaItemPropertyAlbumTitle] = title
+      info[MPMediaItemPropertyArtist] = title
+      info[MPMediaItemPropertyAlbumTitle] = author
       info[MPMediaItemPropertyPlaybackDuration] = current.end - current.start
       info[MPNowPlayingInfoPropertyElapsedPlaybackTime] = chapters.currentElapsedTime(
         currentTime: mediaProgress.currentTime


AudioBooth/AudioBooth/Services/NowPlayingManager.swift
```diff
@@ -175,7 +175,8 @@ final class NowPlayingManager {
 
     if !preferences.showFullBookDuration, let chapters, let current = chapters.current {
       info[MPMediaItemPropertyTitle] = current.title
-      info[MPMediaItemPropertyArtist] = title
```
jeremygrenier, line 178: That is intentional because on the iPhone people prefer to see the book title instead of the artist on their lock screen.

Unfortunately `MPNowPlayingInfoCenter` is not really made for audiobooks so it is a bit tricky. I think what we can do is use  `MPMediaItemPropertyAlbumTitle` for the author instead if you have chapters.

But make sure to run several tests because `MPMediaItemPropertyAlbumTitle` controls a lot of things.

It affects the iPhone Now Playing interface on the Lock Screen and in Control Center, as well as CarPlay and the watchOS companion Now Playing view.

octopotato: oh interesting.. I was comparing audiobooth to how audible does it and there they show the author 🤔

jeremygrenier: I was showing the author before and received several requests to change it. I know apps like BookPlayer and other audiobook apps do the same.

Maybe this could be a user preference, but I feel the app already has a ton of preferences. Maybe just using `MPMediaItemPropertyAlbumTitle` for author in this case will be good enough.

octopotato: yea when I was doing the per book setting I was like "omg so many preferences" lol, I'm also hesitant of making this a preference.. I think `MPMediaItemPropertyAlbumTitle` is a good middle ground for now and see what feedback we get

octopotato: validated.. no noticable change for phone/watch (as expected), carplay gets book title though

PR #184, opened 2026-03-13
Fix crash when tapping author chip

marcosgarrido:
Fixes a crash when tapping the author chip in `BookDetailsView`.

The view was using `NavigationDestination.author`, but `LibraryPageModel`
only supports `NavigationDestination.authorLibrary`, causing a fatalError.

Thanks to @ninewalker on Discord for reporting the issue.


jeremygrenier:
Thank you for making a PR. Even if you adjust prevent the crash that is not the expected behavior. We have a dedicated author page detail.

The issue is I forgot to correctly migrate some of the navigation `.author` to use the new `AuthorDetailsView` instead of `LibraryPage`.

I'm closing the PR and will push the fix.

jeremygrenier:
If you are curious https://github.com/AudioBooth/AudioBooth/commit/0cd6dacc5bfddb42620e3a832b95890324a08cdc.

PR #185, opened 2026-03-14
Add contributor signing setup

octopotato:
I moved all signing settings out of the Xcode project file and into xcconfig files so contributors can override them without touching the pbxproj.

Each target gets its own xcconfig (`Signing.App.xcconfig`, `Signing.Watch.xcconfig`, `Signing.Widget.xcconfig`) that derives from a shared `Signing.xcconfig` base. The base has the current production values as defaults, then does `#include? "Local.xcconfig"` to pull in contributor overrides if they exist. No `Local.xcconfig` = everything works exactly like before.

Contributors run `scripts/setup_contributor` which auto-detects their signing identity, asks for a GitHub username (for the bundle ID), and generates `Local.xcconfig`. The file is gitignored.

I also added stripped-down contributor entitlements that skip capabilities free accounts can't provision (NFC, iCloud KV-store, CarPlay, Wi-Fi info). A `CONTRIBUTOR_BUILD` compile flag disables iCloud sync at compile time so `NSUbiquitousKeyValueStore` doesn't crash without the entitlement. `ModelContextProvider` now falls back to Application Support instead of `fatalError`-ing when the app group container is missing.

For device builds on a free account, contributors still need to manually remove In-App Purchase from Signing & Capabilities because Xcode infers that capability from the StoreKit framework linkage and xcconfig can't control capabilities. Documented in CONTRIBUTING.md.

Simulator builds work with zero setup.


octopotato:
Not sure how valuable this is.. when I tried to test a fix I was doing on one of my devices I had to edit a bunch of the `pbxproj` so I wanted to see how I can reduce the chance of me accidentally checking in a change unintentionally. This was a setup I saw in another project and thought it'd be fun to see if I can get it working for audiobooth.

@jeremygrenier you should double check this doesn't break anything for your flows/app releases 😄

jeremygrenier:
@octopotato I think using a `.xcconfig` makes sense to make contribution simpler, however we should only need the following in a `Signing.xcconfig`:

```
DEVELOPMENT_TEAM = 7MM2M9KD2W
ORG_IDENTIFIER = me.jgrenier
```

We can then use those 2 variables in the per-target `.xcconfig` files, but also directly in the `entitlements` files, so we do not need custom entitlements per contributor.

I would also avoid adding a script to manage this. It is not something that needs to happen often, and adding a bash script to the repo introduces risk I would rather not have to deal with.

AudioBooth/Signing.App.xcconfig
```diff
@@ -0,0 +1,5 @@
+// Signing.App.xcconfig: AudioBooth main app target
+#include "Signing.xcconfig"
+
+PRODUCT_BUNDLE_IDENTIFIER = $(ORG_IDENTIFIER).AudioBS
+CODE_SIGN_ENTITLEMENTS = AudioBooth/AudioBooth.entitlements
```
jeremygrenier: ❓ Any reason we need it now that we use `$(ORG_IDENTIFIER)` in the entitlements?

AudioBooth/Signing.App.xcconfig
```diff
@@ -0,0 +1,5 @@
+// Signing.App.xcconfig: AudioBooth main app target
+#include "Signing.xcconfig"
+
+PRODUCT_BUNDLE_IDENTIFIER = $(ORG_IDENTIFIER).AudioBS
```
jeremygrenier: I'm wondering if we even need one xcconfig by target instead of just using `$(ORG_IDENTIFIER).AudioBS` inline in the project config.

octopotato: good catch! Leftover from when I had the script

Issue #189, opened 2026-03-14
Airplaying to a TV shows a black screen with video controls


b-harting:
## Beta Check
Before opening this issue, please test if the bug is still present in the latest beta version available on [TestFlight](https://testflight.apple.com/join/cAG6dVeN). Investigating issues that are already fixed takes time away from making the app better.

- [x] I have tested on the latest beta and the issue is still present.

## Description
When Airplaying to a TV, the TV plays the audio, but shows a black screen with video controls.

## Steps to Reproduce
1. Start playing an audiobook
2. Airplay to a TV (in my case, a Roku streaming device)
3. See error

## Expected Behavior
Like when airplaying music from Apple Music, I expect to see the Now Playing screen, with the album art, metadata, and have the same forward and back behaviors of the Now Playing screen on the iPhone.

## Actual Behavior
Audio plays, but the screen is black with the video playback seek bar at the bottom.

## Screen Shots

![Image](https://github.com/user-attachments/assets/366cf91f-9334-47b7-83e2-56b257110c1f)
Current Airplay to TV from Audio Booth behavior.

![Image](https://github.com/user-attachments/assets/6593fbd5-4846-4a48-aa66-ed6e5c0d2ce1)
Airplay Expected Behavior

## Device Information
- **Platform:** iPhone (airplaying to a Roku streamer)
- **Device Model:** iPhone 14
- **OS Version:** iOS 26.2
- **AudioBooth Version:** Version: 1.8 (1773371999)
- **Audiobookshelf Server Version:** 2.33.0

jeremygrenier:
Thank you for fixing the issue. #190

PR #190, opened 2026-03-14
Fix for "Airplaying to a TV shows a black screen with video controls" #189

b-harting:
When airplaying audio to a device, if allowsExternalPlayback has not been set to false, the device will assume it is a video stream. If the receiving device is a TV, it  will show a black screen with video controls.  Setting allowsExternalPlayback = false will instead have the TV show the Now Playing view.


jeremygrenier:
Thank you @b-harting for the contribution. I'm not very familiar with `allowsExternalPlayback`, but I'm concerned it might prevent people from using external remotes to control playback, such as using a TV remote to pause playback.

I don't think it's a hard blocker, but if you could do some testing to evaluate the pros and cons of enabling it, that would be great.

Another thing to try is being more explicit with the options in `BookPlayerModel.configureAudioSession`.

Something like:
```swift
try audioSession.setCategory(
  .playback,
  mode: .spokenAudio,
  policy: .longFormAudio,
  options: [.allowAirPlay, .allowBluetoothHFP, .allowBluetoothA2DP]
)
```

b-harting:
I've tested the allowsExternalPlayback version of the code with the TV remote. It still allows for pause, resume play, and skip back and forward (by the skip seconds defined in AudioBooth settings); the same audio controls available from within Now Playing.

I will check with the audioSession options change, as opposed to allowsExternalPlayback, and see if the behavior is the same.

b-harting:
I reverted the code and updated the audioSession options with allowAirPlay, .allowBluetoothHFP, and .allowBluetoothA2DP.  The airplay playback black screen returned. 

The only info I have found in official documentation from Apple on allowsExternalPlayback is "A Boolean value that indicates whether the player allows switching to external playback mode." Unfortunately, what I can't find with that is an official statement of what "external playback mode" is.

Unofficially, I have found the following. On Github [(link)](https://github.com/theos/sdks/blob/master/iPhoneOS16.5.sdk/System/Library/Frameworks/AVFoundation.framework/Headers/AVPlayer.h), I found the header file for AVPlayer, as of iOS 16. Here is the definition there:

```
/*
 @category		
	AVPlayer (AVPlayerExternalPlaybackSupport)
 
 @abstract
	Methods for supporting "external playback" of video 
 
 @discussion
	"External playback" is a mode where video data is sent to an external device for full screen playback at its original fidelity.
	AirPlay Video playback is considered as an "external playback" mode.
 
	In "external screen" mode (also known as mirroring and second display), video data is rendered on the host 
	device (e.g. Mac and iPhone), rendered video is recompressed and transferred to the external device, and the
	external device decompresses and displays the video.
 
	AVPlayerExternalPlaybackSupport properties affect AirPlay Video playback and are the replacement for the 
	deprecated AVPlayerAirPlaySupport properties.
 
	Additional note for iOS: AVPlayerExternalPlaybackSupport properties apply to the Lightning-based
	video adapters but do not apply to 30-pin-connector-based video output cables and adapters.
 */

@interface AVPlayer (AVPlayerExternalPlaybackSupport)

/* Indicates whether the player allows switching to "external playback" mode. The default value is YES. */
@property BOOL allowsExternalPlayback API_AVAILABLE(macos(10.11), ios(6.0), tvos(9.0)) API_UNAVAILABLE(watchos);
```

So, if this is set to false, external playback of video will not happen, but audio continues.

I checked out PocketCast's source for their iOS player [(link)](https://github.com/Automattic/pocket-casts-ios/blob/071808ed44f2a80e6e6777da2d8040ec52934501/podcasts/DefaultPlayer.swift#L679). They are calling the following on initialization of their player:

```
    private func configurePlayer(videoPodcast: Bool) {
        #if !os(watchOS)
            player?.allowsExternalPlayback = videoPodcast
        #endif
```
They are only enabling allowsExternalPlayback if the source is video.

As AudiobookShelf has no support for video podcasts, and based on their tickets [3937](https://github.com/advplyr/audiobookshelf/issues/3937) and [2196](https://github.com/advplyr/audiobookshelf/issues/2196), they are not going to add that support in the near future, I think we can safely set allowsExternalPlayback to false throughout.

b-harting:
Not a problem at all. Just doing due diligence.

PR #206, opened 2026-04-01
Fix playback failure when file codec doesn't match extension

Creationsss:
fix audio files failing to play when the actual codec doesn't match the file extension (e.g. an ogg file named `.mp3`), which caused repeated  "reconnecting..." toasts
for downloaded files, create a temporary symlink with the correct extension derived from the track's actual mimetype/codec so avplayer uses the right decoder
for streaming, force transcode on first recovery attempt instead of retrying direct play which fails identically

only tested on iphone

### DIFF
--- a/API/Sources/API/Models/AudioTrack.swift
+++ b/API/Sources/API/Models/AudioTrack.swift
@@ -1,0 +1,23 @@
+import Foundation
+
+public struct AudioTrack: Codable, Sendable {
+  public let index: Int
+  public let startOffset: Double
+  public let duration: Double
+  public let title: String?
+  public let updatedAt: Date?
+  public let metadata: Metadata?
+  public let format: String?
+  public let bitRate: Int?
+  public let codec: String?
+  public let channels: Int?
+  public let channelLayout: String?
+  public let mimeType: String?
+  public let ino: String
+
+  public struct Metadata: Codable, Sendable {
+    public let filename: String?
+    public let ext: String?
+    public let size: Int64?
+  }
+}
--- a/API/Sources/API/Models/Book.swift
+++ b/API/Sources/API/Models/Book.swift
@@ -53,7 +53,7 @@ extension Book {
   public var duration: Double { media.duration ?? 0 }
   public var size: Int64? { media.size }
   public var chapters: [Media.Chapter]? { media.chapters }
-  public var tracks: [Media.Track]? { media.tracks }
+  public var tracks: [AudioTrack]? { media.tracks }
   public var tags: [String]? { media.tags }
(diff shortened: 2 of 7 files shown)


jeremygrenier:
Thanks for the PR! The codec/extension mismatch is a real issue worth fixing.

However, I think the fix belongs at download time rather than in the player. The `mimeType` and `codec` are already available on the `apiTrack` object when we download, so we can simply use the correct extension when saving the file to disk rather than working around the mismatch at playback time with symlinks.

For example, in `DownloadManager.swift` line 606:
```swift
let trackFile = bookDirectory.appendingPathComponent("\(apiTrack.index)\(ext)")
```

We could derive the correct extension from `apiTrack.mimeType` or `apiTrack.codec` when they disagree with `metadata.ext`, and save the file with the right extension from the start. Same for episodes (line 489).

This avoids adding symlink management to the player and keeps the player focused on playback.

Also, the `forceTranscode: true` change on recovery should be reverted. The original `recoveryAttempts > 1` already forces transcode after the first retry, which handles this case without transcoding on every recovery (e.g. network blips).

Creationsss:
> Thanks for the PR! The codec/extension mismatch is a real issue worth fixing.
> 
> However, I think the fix belongs at download time rather than in the player. The `mimeType` and `codec` are already available on the `apiTrack` object when we download, so we can simply use the correct extension when saving the file to disk rather than working around the mismatch at playback time with symlinks.
> 
> For example, in `DownloadManager.swift` line 606:
> 
> ```swift
> let trackFile = bookDirectory.appendingPathComponent("\(apiTrack.index)\(ext)")
> ```
> 
> We could derive the correct extension from `apiTrack.mimeType` or `apiTrack.codec` when they disagree with `metadata.ext`, and save the file with the right extension from the start. Same for episodes (line 489).
> 
> This avoids adding symlink management to the player and keeps the player focused on playback.
> 
> Also, the `forceTranscode: true` change on recovery should be reverted. The original `recoveryAttempts > 1` already forces transcode after the first retry, which handles this case without transcoding on every recovery (e.g. network blips).

hiya moved the fix to download time and reverted the forceTranscode change

jeremygrenier:
Thank you for your PR. While reviewing it, I realized some cleanup was needed in how we decode tracks for books and episodes coming from the API. I updated both to use a shared `AudioTrack` model. I also added a new `sanitizedExt` computed property to provide the correct file extension, since AVPlayer relies on the extension to determine how to play a file.

Let me know if you have any questions or feedback. Otherwise, I will go ahead and merge the PR. Thank you for your contribution!

PR #211, opened 2026-04-05
Auto-start timer when tapping end of chapter

### LINKED ISSUE #210: [Request] End of Chapter timer should auto start the timer when the button is clicked
I often find myself clicking "end of chapter" timer and it's really annoying that all the other timer buttons auto start but not the end of chapter timer. The app should distinguish between clicking "end of chapter" vs the stepper, the former should auto start while the latter should wait until "start" or "end of chapter" is clicked again to confirm/start.

octopotato:
- Closes https://github.com/AudioBooth/AudioBooth/issues/210

All the preset timer buttons (15 min, 30 min, etc.) auto-start the timer when tapped, but "End of chapter" required a separate tap on the Start button. Since I use "End of chapter" the most, it was really annoying to have to click 2 buttons to start end of chapter timers.

I added an `autoStart` param to `onChaptersChanged` so tapping the main row starts the timer immediately (same flow as `onQuickTimerSelected`), but the +/- steppers still just adjust the count. If you bump the stepper to 3 chapters then tap the row, it honors that value instead of resetting to 1.

https://github.com/user-attachments/assets/8c49720c-c04e-4bfe-9ce4-f9116d0872a4

### DIFF
--- a/AudioBooth/AudioBooth/Screens/BookPlayer/Sheets/Timer/TimerPickerSheet.swift
+++ b/AudioBooth/AudioBooth/Screens/BookPlayer/Sheets/Timer/TimerPickerSheet.swift
@@ -160,7 +160,10 @@ struct TimerPickerSheet: View {
     }()
 
     HStack {
-      Button(action: { model.onChaptersChanged(chapterCount, autoStart: true) }) {
+      Button(action: {
+        model.onChaptersChanged(chapterCount)
+        model.onStartTimerTapped()
+      }) {
         VStack(alignment: .leading, spacing: 2) {
           Text(chapterCount == 1 ? "End of chapter" : "End of \(chapterCount) chapters")
             .font(.system(size: 16, weight: .medium))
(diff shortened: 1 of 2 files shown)


AudioBooth/AudioBooth/Screens/BookPlayer/Sheets/Timer/TimerPickerSheet.swift
```diff
@@ -160,7 +160,7 @@ struct TimerPickerSheet: View {
     }()
 
     HStack {
-      Button(action: { model.onChaptersChanged(1) }) {
+      Button(action: { model.onChaptersChanged(chapterCount, autoStart: true) }) {
```
jeremygrenier: I think it will be cleaner to do

```suggestion
      Button(action: { 
        model.onChaptersChanged(chapterCount)
        model.onStartTimerTapped()
      }) {
```

instead of passing a `autoStart` or have a new function `onChaptersTapped(chapterCount)` what will call `onChaptersChanged` and `onStartTimerTapped` under the hood.

octopotato: yea that's much simpler.. good call!

jeremygrenier (approved):

PR #213, opened 2026-04-05
Add reset all progress button to collection views

octopotato:
I wanted a way to reset progress for a whole series at once instead of going through each book's context menu individually. Useful for relistens.

I added "Reset All Progress" and "Mark All as Finished" to the ellipsis menu in series views. Both reuse the existing context menu actions (`onResetProgressTapped` / `onMarkAsFinishedTapped`) and only show when relevant (at least one book has progress / is unfinished).

https://github.com/user-attachments/assets/d32f75dc-4805-4749-b7d5-591d1fa0b9c3


AudioBooth/AudioBooth/Screens/Library/Library/LibraryPage.swift
```diff
@@ -84,6 +84,22 @@ struct LibraryPage: View {
...
+              message:
+                "This will clear all listening progress for all books in this collection. This cannot be undone.",
+              action: "Reset All"
+            ),
+            action: model.onResetAllProgressTapped
+          ) {
+            Label("Reset All Progress", systemImage: "arrow.counterclockwise")
+          }
+          .tint(.primary)
+        }
```
jeremygrenier: Do you mind moving it to the ellipsis menu like for `BookDetailsView`? Also like `BookDetailsView` we should make sure to show it only if at least one book has some progress.

I also think it could be great to have a **Mark as Finished** option as well if not all the books are completed.

jeremygrenier: I also think we should only show it when the filtering is series. It doesn't make much sense to have it for authors or narrators or tags.

AudioBooth/AudioBooth/Screens/Library/Library/LibraryPageModel.swift
```diff
@@ -178,6 +178,24 @@ final class LibraryPageModel: LibraryPage.Model {
...
+        if let localBook = try? LocalBook.fetch(bookID: model.id) {
+          try? await localBook.resetProgress()
+        } else {
+          let book = try? await audiobookshelf.books.fetch(id: model.id)
+          try? await book?.resetProgress()
+        }
+
+        model.cover.progress = 0
+      }
+    }
```
jeremygrenier: The logic is quite over complicated.


```suggestion
    for case let .book(model) in items {
      model.contextMenu?.onResetProgressTapped()
    }
```

AudioBooth/AudioBooth/Screens/Library/Library/LibraryPageModel.swift
```diff
@@ -44,6 +44,7 @@ final class LibraryPageModel: LibraryPage.Model {
       super.init(
         hasMorePages: true,
         isRoot: false,
+        showSeriesActions: true,
```
jeremygrenier: Instead of having `showSeriesActions` then computed var in the view. Do you mind adjusting it to follow the `BookDetailsView` behavior?

You can define

```swift
    struct Actions: OptionSet {
      let rawValue: Int

      static let markAsFinished = Actions(rawValue: 1 << 0)
      static let resetProgress = Actions(rawValue: 1 << 1)
    }
    var actions: Actions
```

then the view model will be the one deciding which actions should be visible.

jeremygrenier (changes requested):

AudioBooth/AudioBooth/Screens/Library/Library/LibraryPageModel.swift
octopotato: done, let me know what you think

octopotato:
only thing I don't really like over what the computed var is that the menu items aren't always accurate but I think it's an acceptable trade off. Ex

https://github.com/user-attachments/assets/045f4b58-a478-483d-a338-cae46ab45fe5

jeremygrenier:
The issue is `model.cover.progress` doesn't update on appear. You can use `MediaProgress.progress(for: bookID)` which should be more accurate.

I should probably also take a look why `model.cover.progress` is not correctly updating but I still think using `MediaProgress.progress` is cleaner.

octopotato:
nice that did the trick

https://github.com/user-attachments/assets/9473f564-e8e3-4c76-a7c7-32a1a06ac2a0

jeremygrenier (approved):

Issue #237, opened 2026-05-05
Custom headers are not sent to /public/session/* endpoints


mdarvanaghi:
## Beta Check
Before opening this issue, please test if the bug is still present in the latest beta version available on [TestFlight](https://testflight.apple.com/join/cAG6dVeN). Investigating issues that are already fixed takes time away from making the app better.

- [x] I have tested on the latest beta and the issue is still present.

## Description
When the app accesses e.g. the /public/session/\<GUID>/track/\<NUMBER> endpoints, the custom headers configured for the server are not sent with the request. For Audiobookshelf instances hosted behind zero-trust providers such as Pangolin or Cloudflare, playback breaks without whitelisting that path entirely.

Here's an example from my Pangolin request logs:

<img width="802" height="516" alt="Image" src="https://github.com/user-attachments/assets/9a883a63-770f-42d1-a7b9-6cd923d91592" />

## Steps to Reproduce
1. Open AudioBooth from a non-whitelisted IP (such as on mobile data)
2. Select server which is setup behind zero-trust reverse proxy with custom access headers
3. Play any audiobook which is not downloaded
4. Notice "Reconnecting" error and no playback

## Expected Behavior
Custom headers are sent with all requests and playback works

## Actual Behavior
Playback does not work

## Screenshots
<img width="585" height="1266" alt="Image" src="https://github.com/user-attachments/assets/97925001-eccf-49fd-ba71-1289c8043dd0" />

## Device Information
- **Platform:** iPhone
- **Device Model:** 12
- **OS Version:** 26.4.2
- **AudioBooth Version:** 1.9 and 1.10
- **Audiobookshelf Server Version:** 2.34

jeremygrenier:
Hi, this is by design. The `public/` path is meant to be publicly accessible.

If you want more context, see https://github.com/AudioBooth/AudioBooth/issues/90#issuecomment-3746870595.

Also, regarding your PR: `AVURLAssetHTTPHeaderFieldsKey` is a private API and not publicly available, so using it can lead to rejection from Apple during App Review.

mdarvanaghi:
I can't find documentation on the /public/ endpoints, so it's hard to tell the intention with them. I agree that 'public' sounds like they should be accessible without (ABS) user authorization. But still I would prefer to be able to protect my Audiobookshelf instance from unwanted traffic, on any endpoint, hence the custom headers for access.

[...]

PR #238, opened 2026-05-05
Add custom headers to track requests

mdarvanaghi:
Addresses this issue: https://github.com/AudioBooth/AudioBooth/issues/237

I unfortunately do not own a mac so I cannot test this fix.


jeremygrenier:
Hi, as I mentioned on the related issue, I'm going to pass on this.

`AVURLAssetHTTPHeaderFieldsKey` is a private API, not publicly available, and Apple explicitly recommends against using it: https://developer.apple.com/forums/thread/20421?answerId=808239022#808239022. Shipping it risks App Store rejection.

The public alternative is `AVAssetResourceLoader` with a custom URL scheme, but it is far from trivial to implement correctly (byte-range requests, content info, HLS segment handling, etc.), and that's a lot of complexity to add for content served from the `public/` path, which is intended to be publicly accessible by design.

Closing this one. Thanks for the contribution.

Issue #255, opened 2026-05-26
[Bug] Switching from "Playlists" to "Collections" on iOS creates a weird visual effect


calvinbrown085:
## Beta Check
Before opening this issue, please test if the bug is still present in the latest beta version available on [TestFlight](https://testflight.apple.com/join/cAG6dVeN). Investigating issues that are already fixed takes time away from making the app better.

- [ ] I have tested on the latest beta and the issue is still present.

## Description
Switching from "Playlists" to "Collections" on iOS creates a weird visual effect

This creates a weird visual effect where playlists appears off the screen for a bit and moves back. 
## Steps to Reproduce
1. Go to Collections
2. Tap on Playlists
3. Tap on Collections

## Expected Behavior
Slide over like going from Collections to Series 

## Actual Behavior
Weird visual effect bug

## Screenshots
https://github.com/user-attachments/assets/903657c5-2941-4a10-abd4-48c6082490ed

## Device Information
- **Platform:** iPhone
- **Device Model:** iPhone 16 Pro Max
- **OS Version:** iOS 26.5
- **AudioBooth Version:** 1.9
- **Audiobookshelf Server Version:** (if known)

jeremygrenier:
Hello thank you for the report, the issue should already be fixed in the beta if you want to give it a try.

calvinbrown085:
@jeremygrenier Oh cool! I noticed the code wasn’t fixed on main, am I looking in the wrong place?

jeremygrenier:
I fixed the issue in [this commit](https://github.com/AudioBooth/AudioBooth/commit/4f410a70cfbf85c62cef145e7fcc522327aecebc). It does a bit more than just fixing the issue, since I also did some small cleanup at the same time, but `.fixedSize(horizontal: true, vertical: true)` is what prevented the animation issue.

calvinbrown085:
@jeremygrenier Ah cool. When I ran in the simulator it must have already been fixed! 

Thanks for taking the time to look at those

PR #256, opened 2026-05-26
Fixing visual UI bug collections switch

calvinbrown085:
Fix for: https://github.com/AudioBooth/AudioBooth/issues/255

It tested this in the iOS simulator and seems to have fixed the bug!

https://github.com/user-attachments/assets/9160ca3a-98d9-4ebd-ba82-93e4c532ee6f

Thanks!


jeremygrenier:
Hello thank you for the report, the issue should already be fixed in the beta if you want to give it a try.

PR #264, opened 2026-06-02
feat: add alarm

DonQuinleone:
## Summary and use-cases

Adds a new Alarm feature, which makes it possible to play an alarm at a set time or after a specified duration has passed. This is designed to complement the existing Timer feature, not replace it, providing a solution for different scenarios.

For instance, the timer might be used when going to bed, quietly disappearing without requiring user intervention. Whereas, the alarm is designed to interrupt the listening experience and require user intervention to turn it off; it could be used to ensure that you wake up when you are on the train, to remind you to take medication, to signify the end of your break from work, etc.

Of course, the iOS Clock app can be used to trigger alarms, but this has one significant shortcoming - it will play out of both the selected playback device and the device's speaker. This makes it unsuitable for use in environments where the alarm must only come through headphones, for example.

I was on the train in the early hours of today, and was a little paranoid that I would fall asleep and miss my connection. Having this feature would have allowed me to listen to my audiobook of choice whilst ensuring that I did not miss my stop!

image
image
image
image
image

### Detailed implementation overview

* Adds a new `Alarm` button to the player, enabled by default, but hideable via Controls & Layout in the settings.
* New alarm can be set with provided duration or a specific time.
* More time can be added to the alarm at will, which will delay the alarm going off by a selected amount.
* Prior to the alarm going off, the audiobook will be faded out over a duration specified in the settings - fading can also be disabled completely.
* When an alarm is triggered, the user will be sent a notification.
* When an alarm is active, the button on the player will be lit accordingly. For parity, this has also been added to the Timer, so the user can quickly see when the timer is active.

I have done my best to follow the style and idioms of the codebase.

Please let me know if you have any feedback. Otherwise, I hope you like this addition.


AudioBooth/AudioBooth/Screens/BookPlayer/Sheets/Alarm/AlarmPickerSheetModel.swift
```diff
@@ -0,0 +1,181 @@
+import AVFoundation
+import Foundation
+
+final class AlarmPickerSheetViewModel: AlarmPickerSheet.Model {
+  private let preferences = UserPreferences.shared
+  private let itemID: String
+  private let player: AudioPlayer
+  private weak var timerViewModel: TimerPickerSheetViewModel?
```
jeremygrenier: `AlarmPickerSheetViewModel` should not need the `timerViewModel`. It seems to only be used for `applyFadeOut`.

Either you duplicate the code in this VM or even better move the `fadeOut` to `AudioPlayer`.

AudioBooth/AudioBooth/Models/PlayerControl.swift
```diff
@@ -33,6 +36,6 @@ enum PlayerControl: String, CaseIterable, Identifiable, Codable {
   }
 
   static var `default`: [PlayerControl] {
-    [.speed, .timer, .bookmarks]
+    [.speed, .timer, .alarm, .bookmarks]
```
jeremygrenier: Please do not add it as part of the default. I'm trying to keep it simple. People can decide to add it later if they want to.

AudioBooth/AudioBooth/Screens/BookPlayer/Sheets/Timer/TimerPickerSheetModel.swift
```diff
@@ -265,10 +265,10 @@ final class TimerPickerSheetViewModel: TimerPickerSheet.Model {
     }
   }
 
-  private func fadeOut(_ seconds: TimeInterval) {
-    let fadeOut = preferences.timerFadeOut
-    if fadeOut > 0, seconds < fadeOut {
-      player.volume = Float(seconds / fadeOut) * Float(preferences.volumeLevel)
+  func applyFadeOut(secondsRemaining: TimeInterval, fadeDuration: TimeInterval? = nil) {
```
jeremygrenier: No reason to rename the function. Also as stated before we should not expose this function.

AudioBooth/AudioBooth/Screens/Settings/Preferences/Player/SleepShake/SleepPreferencesView.swift
```diff
@@ -118,6 +139,18 @@ struct SleepPreferencesView: View {
...
   }
+
+  private func alarmFadeLabel(_ value: Double) -> String {
+    switch value {
+    case 0: return "No Fade"
+    case 5: return "5 Seconds"
+    case 10: return "10 Seconds"
+    case 15: return "15 Seconds"
+    case 30: return "30 Seconds"
+    case 60: return "1 Minute"
```
jeremygrenier: Why not using ` Duration.seconds(value).formatted` for those values as well? Because currently localization will not work on those strings.

AudioBooth/AudioBooth/Services/UserPreferences/UserPreferences.swift
```diff
@@ -61,6 +61,12 @@ final class UserPreferences: ObservableObject {
   @AppStorage("timerFadeOut")
   var timerFadeOut: Double = 30.0
 
+  @AppStorage("customAlarmDurationMinutes")
+  var customAlarmDurationMinutes: Int = 15
```
jeremygrenier: Do we really need `customAlarmDurationMinutes`? I'm not totally against it but not sure if keeping the last alarm duration provide much value.

AudioBooth/AudioBooth/Screens/BookPlayer/Sheets/Alarm/AlarmPickerSheetModel.swift
```diff
@@ -0,0 +1,181 @@
...
+    let remaining = activeAlarm.nextTrigger.timeIntervalSinceNow
+
+    if remaining <= 0 {
+      countdownText = "00:00:00"
+      player.pause()
+      playAlarmTone()
+      stopMonitoring()
+      return
+    }
+    countdownText = formatCountdown(remaining)
```
jeremygrenier: Instead of using `countdownText` I think you should use SwiftUI.Text function to render timers.

Either https://developer.apple.com/documentation/swiftui/text/datestyle/relative or https://developer.apple.com/documentation/swiftui/text/init(timerinterval:pausetime:countsdown:showshours:)

AudioBooth/AudioBooth/Screens/EbookReader/EbookReaderViewModel.swift
```diff
@@ -439,7 +439,10 @@ final class EbookReaderViewModel: EbookReaderView.Model {
     var location = navigator?.currentLocation
     location?.locations.totalProgression = nil
 
-    let ebookLocation = location?.jsonString
+    let ebookLocation: String? = {
+      guard let location else { return nil }
+      return try? location.jsonString()
+    }()
```
jeremygrenier: Why do we need that?

AudioBooth/AudioBooth/Services/AlarmNotificationService.swift
```diff
@@ -0,0 +1,60 @@
...
+    if settings.authorizationStatus == .authorized || settings.authorizationStatus == .provisional {
+      return true
+    }
+
+    if settings.authorizationStatus == .denied {
+      return false
+    }
+
+    do {
+      return try await center.requestAuthorization(options: [.alert, .sound, .badge])
```
jeremygrenier, line 22: You will need to add `NSUserNotificationsUsageDescription` to the `AudioBooth/AudioBooth/Info.plist` for the permission request to work.

AudioBooth/AudioBooth/Screens/BookPlayer/Sheets/Alarm/AlarmPickerSheet.swift
```diff
@@ -0,0 +1,182 @@
...
+                  get: { Double(model.addTimeMinutes) },
+                  set: { model.onAddTimeMinutesChanged(Int($0.rounded())) }
+                ),
+                in: 1...30,
+                step: 1
+              )
+            }
+
+            HStack(alignment: .center, spacing: 10) {
+              let cancelButtonTitle = model.countdownText == "00:00:00" ? "Stop" : "Cancel"
```
jeremygrenier: `model.countdownText == "00:00:00"` to decide if it should be stop or cancel is not great. We should avoid using string comparison for that.

AudioBooth/AudioBooth/Services/AlarmNotificationService.swift
```diff
@@ -0,0 +1,60 @@
...
+    } catch {
+      return false
+    }
+  }
+
+  func schedule(itemID: String, triggerDate: Date) async {
+    cancel(itemID: itemID)
```
jeremygrenier: I think we should probably make sure to cancel all the `playerAlarm` notification no? Not even sure we should have different id for different notification by item ID. Probably better to make sure we only have one alarm at the time.

AudioBooth/AudioBooth/Services/AlarmNotificationService.swift
```diff
@@ -0,0 +1,60 @@
+import Foundation
+import UserNotifications
+
+@MainActor
+final class AlarmNotificationService {
+  static let shared = AlarmNotificationService()
```
jeremygrenier: Do we really need the shared? Can we just use `AlarmNotificationService()` in our view model?

jeremygrenier (commented):
Thank you for the contribution and the PR. I have several pieces of feedback.

Also, I'm not sure whether we should have a dedicated Alarm sheet or update the existing Timer sheet to support alarms as well. It might make more sense to make alarms part of the Timer feature, perhaps through a sub-tab or a simple toggle.

AudioBooth/AudioBooth/Models/PlayerControl.swift
DonQuinleone: Removed the .alarm PlayerControl case and reverted default controls to speed, timer, bookmarks.

AudioBooth/AudioBooth/Screens/BookPlayer/Sheets/Alarm/AlarmPickerSheetModel.swift
DonQuinleone: Removed AlarmPickerSheet / AlarmPickerSheetViewModel and moved into the Timer feature.

AudioBooth/AudioBooth/Screens/BookPlayer/Sheets/Timer/TimerPickerSheetModel.swift
DonQuinleone: Restored

AudioBooth/AudioBooth/Services/UserPreferences/UserPreferences.swift
DonQuinleone: Yes, true. Removed.

AudioBooth/AudioBooth/Screens/EbookReader/EbookReaderViewModel.swift
DonQuinleone: Thanks for flagging this.
I needed this because in my current setup location?.jsonString was being treated as a function reference rather than a String?, which caused a type mismatch at the call site.
So I made it explicit with try? location.jsonString() to ensure we pass the expected String?.
If there’s a cleaner way to handle this, I’m very happy to update it.

AudioBooth/AudioBooth/Services/AlarmNotificationService.swift
DonQuinleone: Good point - done.

DonQuinleone:
Thanks so much for your feedback! I've added a new commit that address each of your points, including moving the Alarm into the Timer sheet.
image
Let me know if you need any further amends :-)

loeffner:
I feel like this feature goes against the following statement in the readme:

„AudioBooth focuses on what matters most: enjoying your audiobooks. No unnecessary features, no bloat…“

A timer is already available in iOS and I am sure there are several apps that can do what you want. 

But I’m just a user, so of course this is completely up to the maintainer. I just wanted to provide my pov.

jeremygrenier:
> I feel like this feature goes against the following statement in the readme:
> 
> „AudioBooth focuses on what matters most: enjoying your audiobooks. No unnecessary features, no bloat…“
> 
> A timer is already available in iOS and I am sure there are several apps that can do what you want.
> 
> But I’m just a user, so of course this is completely up to the maintainer. I just wanted to provide my pov.

I mostly agree with you. I was on the fence about pushing back on it or not. It’s always hard to find the right balance between satisfying users and avoiding too many unnecessary features.

If this is hidden within the Timer sheet, it probably won’t affect most users. However, it still increases the complexity of the codebase and makes the app harder to maintain.

DonQuinleone:
It's, of course, entirely up to you.

If you'd prefer not to merge in, I won't be offended - I'll just maintain my own fork, because for me it will be very useful.

jeremygrenier:
@DonQuinleone I did some cleanup and polish the timer sheet to properly handle both. I also update the dependencies on `main` and rebase the branch. Please give it a try and let me know if it works for you.

Also after doing it I just recall Apple released https://developer.apple.com/documentation/AlarmKit last year. I'm wondering if it will be better to use it instead of local notification. I never played with it so not sure what is possible or not.

DonQuinleone:
Looks superb - thanks so much for working with me on this.
I'd be really happy with this as-is.

As for using AlarmKit, I'm honestly not sure - I haven't used it before either. This is always the way that I've done things in the past. I can look into it though, if that would be useful.

jeremygrenier (approved):
Thank you for your contribution

PR #269, opened 2026-06-05
Add Share button for downloaded ebooks and audiobooks

antdon:
Hi, I like your app. I've been trying to use it for ebooks and audiobooks and found that I often wanted to open ebooks in the apple books app. For that purpose I've implemented share functionality. 

* When you download a book (audio book or ebook) the download button becomes a share button
* The share button will open the share menu and allow you to share the file
* Apple handles this nicely so dependent on format it allows you to send via imessage or to kindle or to apple books
* When you delete the book from your downloads the share button reverts to the download button
* If you have the audio book and ebook you get a drop down menu to choose which one to download
* If you have multiple formats for your ebook like pdf, epub and mobi it will download the primary one
* If you have an audiobook with many tracks it will zip them before sharing, this can take a second so we have the share button become a spinner while we are zipping

Here are some images to show you what it looks like on iPad 11" and iphone13. I didn't include a screenshot of the share screen since it looks the same as on every other app.

IMG_0008 (1)
IMG_0008
IMG_0005
IMG_0006 (1)
IMG_0006
IMG_0007


jeremygrenier:
Thanks for the PR and the kind words about the app!

This is actually already possible today via the "Open on Web" option in the ellipsis menu. From there you can open the file in any app you'd like, including Apple Books. You can also long-press the file in the ebook section of the page to get the same options. Maybe we should rename it to "Share", though that's not quite the right wording either.

I'm not a huge fan of having this so prominent on the book details page, as it's an edge case for most users.

antdon:
When I click open on web I get unauthorized. I assume it needs an API key somewhere? I'm not sure what that API does but you're still opening up a webview and making an ABS api call for a file that you have downloaded. Apple's share works really nicely here. I'm happy to move it into ... menu or wherever you think is more appropriate.

jeremygrenier:
I think you are right depending of how your access your server you may have permissions issue. Do you mind adjusting the PR to instead replace "Open on Web" to be the share behavior?

AudioBooth/AudioBooth/Screens/BookDetails/BookDetailsViewModel.swift
```diff
@@ -625,16 +659,120 @@ final class BookDetailsViewModel: BookDetailsView.Model {
 }
 
 extension BookDetailsViewModel {
-  private func openEbookInSafari(_ url: URL) {
-    let safariViewController = SFSafariViewController(url: url)
-    safariViewController.modalPresentationStyle = .overFullScreen
-
-    if let windowScene = UIApplication.shared.connectedScenes.first as? UIWindowScene,
+  private func presentShareSheet(for fileURL: URL) {
```
jeremygrenier: Any reason to not use SwiftUI `ShareLink`?

AudioBooth/AudioBooth/Screens/BookDetails/BookDetailsViewModel.swift
```diff
@@ -625,16 +659,120 @@ final class BookDetailsViewModel: BookDetailsView.Model {
...
+
+    do {
+      try FileManager.default.copyItem(at: ebookURL, to: destination)
+      return destination
+    } catch {
+      AppLogger.viewModel.error(
+        "Failed to prepare ebook for sharing: \(error.localizedDescription)"
+      )
+      return ebookURL
+    }
```
jeremygrenier: Why do you copy the book to a temporary location? You should be able to share it from the current location no?

AudioBooth/AudioBooth/Screens/BookDetails/BookDetailsViewModel.swift
```diff
@@ -625,16 +659,120 @@ final class BookDetailsViewModel: BookDetailsView.Model {
...
+        let ext = source.pathExtension.isEmpty ? "mp3" : source.pathExtension
+        let trackName = String(format: "%03d.%@", index + 1, ext)
+        try fm.copyItem(at: source, to: folder.appendingPathComponent(trackName))
+      }
+    } catch {
+      return nil
+    }
+
+    let zipDestination = directory.appendingPathComponent("\(name).zip")
+    guard zip(directory: folder, to: zipDestination) else { return nil }
```
jeremygrenier: I do not think AudioBooth should be responsible for zipping the tracks. We should just share one or multiple track files directly.

jeremygrenier (approved):
I address the feedback. Thank you for your contribution.

PR #277, opened 2026-06-15
Refresh the access token on 401 instead of forcing a re-login after a server outage

### LINKED ISSUE #276: App forces full re-login instead of silently reconnecting with the refresh token after a server outage
## Beta Check

This is a logic bug in the shared networking/auth layer, observed on the current App Store release (1.10) and confirmed by source analysis against `main` — the affected code paths are unchanged on the latest beta. Reproduction does not depend on app version or device.

- [x] Confirmed present on the latest source (`main`).

## Description

When the Audiobookshelf server becomes unreachable for a while and then comes back, AudioBooth does **not** silently reconnect. Instead it drops into an authentication-error state and requires the user to re-enter their **username and password**, even though a valid (30-day) refresh token is stored on the device. Other clients (e.g. the official Audiobookshelf app) reconnect automatically in this situation without any user interaction.

This matters because Audiobookshelf **access tokens expire after 1 hour by default** (refresh tokens last 30 days and are meant to mint new access tokens transparently — see [ABS JWT auth discussion #4460](https://github.com/advplyr/audiobookshelf/discussions/4460)). Any outage longer than the access-token lifetime triggers this.

## Steps to Reproduce

1. Sign in to a server (ABS 2.26.0+, JWT auth) and use the app normally.
2. Take the server offline for longer than the access-token lifetime (≈1 hour by default).
3. Bring the server back online.
4. Open / return to AudioBooth and let it talk to the server again.

## Expected Behavior

- The app silently uses the stored refresh token to obtain a new access token and reconnects on its own.
- A simple unreachable-server period shows a **connection** problem (and recovers automatically), never an **authentication** problem.
- No username/password re-entry is required as long as the refresh token is still valid.

## Actual Behavior

- The server indicator turns **red** (`.authenticationError`), not orange (`.connectionError`).
- The app does not retry on its own after the server returns (on a stable Wi-Fi connection nothing re-probes).
- Server → "Server Details" shows "Authentication Required", and the only offered recovery is a full username/password (or API key / OIDC) re-login.

Four independent gaps in the API/auth layer combine to produce this:

1. **No refresh-on-401 + retry.** `NetworkService` treats any HTTP 401 as terminal and sets `.authenticationError` immediately (`API/Sources/API/NetworkService.swift`, ~L201). It never attempts a token refresh and retry. Token refresh is only done *proactively* (by expiry) in `CredentialsActor.freshCredentials`, never *reactively* on a rejected request.

2. **Requests are sent without an `Authorization` header when a proactive refresh can't complete.** The header provider swallows the failure and returns no headers (`API/Sources/API/Audiobookshelf.swift`, ~L64: `let freshToken = try? await server.freshToken; guard let credentials = freshToken else { return [:] }`). Against a *recovered* server, an unauthenticated request returns 401 → the app concludes the credentials are bad, when in reality the (valid) refresh token simply wasn't used.

3. **No automatic reconnection.** The only automatic server probes are at cold launch (`checkServersHealth()` in `AppDelegate`) and on a Wi-Fi/cellular **interface change** that only retries `.connectionError` (`AudioBooth/.../Services/NetworkMonitor.swift`). If the phone stays on the same network while the *server* restarts, nothing re-probes; foregrounding only syncs playback sessions.

4. **Recovery UI only offers full re-login.** The "Authentication Required" path builds an `AuthenticationViewModel` that calls `login(username:password:)` (`AudioBooth/.../Servers/Authentication/AuthenticationViewModel.swift`). It never attempts to recover silently via the stored refresh token.

## Device Information

- **Platform:** iPhone (logic is shared across all platforms, including iPad / Apple Watch)
- **OS Version:** Not device-specific (bug is in the shared networking layer)
- **AudioBooth Version:** 1.10 (and current `main`)
- **Audiobookshelf Server Version:** Any 2.26.0+ (JWT access/refresh tokens)

## Additional Context

A fix PR follows this issue. The fix adds the standard refresh-interceptor behavior (refresh once on 401 and retry), stops sending unauthenticated requests / mis-classifying connection failures as auth failures, and reconnects on foreground — so an outage longer than the access-token lifetime recovers silently using the existing refresh token.

Related upstream context:
- [New JWT authentication · advplyr/audiobookshelf Discussion #4460](https://github.com/advplyr/audiobookshelf/discussions/4460)
- [Duplicate refresh tokens across sessions can cause unexpected logout · advplyr/audiobookshelf #5253](https://github.com/advplyr/audiobookshelf/issues/5253)

spawvn:
## Summary

Fixes #276.

After the Audiobookshelf server is unreachable for longer than the access-token lifetime (access tokens expire after **~1 hour** by default; refresh tokens last 30 days), AudioBooth would drop into an **authentication-error** state and require a full **username/password re-login** when the server came back — instead of silently refreshing the access token with the stored refresh token and reconnecting, the way the official client does.

This PR makes the client recover the way the server's JWT model intends: refresh on a rejected request, retry once, and only ask for credentials when the refresh token itself is genuinely rejected.

## Root cause

Four independent gaps combined to turn a routine token expiry into a forced logout:

1. **No refresh-on-401.** `NetworkService` treated any HTTP 401 as terminal and set `.authenticationError` immediately — token refresh only ever happened *proactively* (by expiry), never *reactively* on a rejected request.
2. **Unauthenticated requests.** When a proactive refresh couldn't run (e.g. the server was unreachable), the auth-header provider returned **no** `Authorization` header. Against a *recovered* server that yields a 401, which was then misread as bad credentials.
3. **Aggressive recovery.** A single 401/403 (including one provoked by gaps 1–2) flipped the server to `.authenticationError`, whose only UI affordance is a full re-login — the valid refresh token was never used.
4. **No reconnect.** The only automatic probes were at cold launch and on a Wi-Fi/cellular interface change (for `.connectionError` only); a server that recovered on the same network was never re-probed.

## What this changes

- **`NetworkService`** — on a 401 for a refreshable connection, refresh the token once and retry the request a single time. Only a *definitive* refresh-token rejection (401/403 from `/auth/refresh`) becomes `.authenticationError`; any other failure is treated as `.connectionError`, so a transient hiccup never forces a re-login. (Bounded to one retry via an `allowAuthRetry` flag — no recursion.)
- **`Audiobookshelf` / `AuthenticationService.verifyAlternativeURL`** — when a proactive refresh can't run, fall back to the stored token instead of sending an unauthenticated request, so a recovered server returns a *recoverable* 401 rather than a misclassified auth failure.
- **`CredentialsActor`** — add a forced refresh that bypasses the expiry check and connection backoff (a 401 proves the server is reachable, so the connection backoff no longer applies). Concurrent 401s **coalesce onto a single in-flight refresh**, so the rotating refresh token is exchanged only once (avoids the rotation race in advplyr/audiobookshelf#5253).
- **`AudioBoothApp`** — when returning to the foreground and not currently connected, re-probe the server (`checkServersHealth()`) so a server that recovered while backgrounded reconnects without user action.
- **`CHANGELOG.md`** — note under `[Unreleased]`.

## Behavior after the fix

- **Outage shorter or longer than the token lifetime, server back, app interacts:** the stored refresh token mints a new access token and the connection recovers silently (green). No re-login.
- **Open the app while the server is still down (e.g. 12h):** requests fail with network errors, so the server shows a **connection** problem (orange), never an **authentication** problem (red); the refresh token is preserved; it reconnects automatically once the server returns.
- **Re-login is still required only** when a reachable server definitively rejects the refresh token (genuinely expired after 30 days, or revoked/logged-out server-side).

## Testing & confidence — please read

In the interest of transparency: **this change was developed from static source analysis and has not been run end-to-end on a device or simulator.**

Verified:
- **The full app compiles cleanly, natively.** `xcodebuild` reports `** BUILD SUCCEEDED **` for the `AudioBooth` scheme on the **iOS 26.5 Simulator with the embedded watchOS 26.5 Watch app** (iOS app + Watch app + widget extensions). All changed files compiled for both the iOS and watchOS architectures — **zero errors, zero warnings**. (Also confirmed independently via a Mac Catalyst build and a standalone `swift build` of the `API` package.)
- `swift-format` passes on all changed files.

Not verified:
- No live run, so the runtime refresh/reconnect behavior has not been observed end-to-end — that needs an Audiobookshelf server that can be taken offline past the access-token lifetime (~1h) and brought back.

Confidence is high because the change is the standard "refresh interceptor" pattern, is surgical (~85 lines, no change to the happy path), preserves existing behavior for API-key/legacy tokens (correctly excluded from the refresh-retry), and was reasoned through the key scenarios above. A maintainer confirming a real expiry → recovery cycle on a device would be a valuable final check before merge.

Not included (kept out of scope intentionally): any user-facing "connection lost" toast/notification, and aggressive background polling while the app stays foregrounded during an outage (recovery there happens on the next request or on foreground).

### DIFF
--- a/API/Sources/API/Audiobookshelf.swift
+++ b/API/Sources/API/Audiobookshelf.swift
@@ -61,8 +61,13 @@ public final class Audiobookshelf: @unchecked Sendable {
         let server = self.authentication.server
       else { return [:] }
 
-      let freshToken = try? await server.freshToken
-      guard let credentials = freshToken else { return [:] }
+      // Prefer a freshly refreshed token. If a proactive refresh can't run (e.g.
+      // the server is unreachable), fall back to the stored token so the request
+      // is still authenticated. A recovered server then answers 401, which
+      // NetworkService recovers from by refreshing and retrying — instead of
+      // receiving an unauthenticated request and reporting an authentication
+      // error for what is really a connection issue.
+      let credentials = (try? await server.freshToken) ?? server.token
 
       var headers = server.customHeaders
       headers["Authorization"] = credentials.bearer
--- a/API/Sources/API/CredentialsActor.swift
+++ b/API/Sources/API/CredentialsActor.swift
@@ -44,23 +44,51 @@ actor CredentialsActor {
         )
       }
 
-      let task = Task<Credentials, Error> { @MainActor [server] in
-        try await Audiobookshelf.shared.authentication.refreshToken(for: server)
-      }
+      return try await performRefresh(for: server)
+    }
+  }
 
-      refreshTask = task
+  /// Forces a refresh of the access token using the stored refresh token,
+  /// ignoring both the expiry check and the connection-error backoff.
+  ///
+  /// This is the recovery path for a 401 on a request we believed was
+  /// authenticated: the rejection proves the server is reachable, so the
+  /// backoff that was set while it was unreachable no longer applies.
+  /// Concurrent callers coalesce onto a single in-flight refresh so the
+  /// rotating refresh token is only exchanged once.
+  func forceRefresh() async throws -> Credentials {
+    if let refreshTask {
+      return try await refreshTask.value
+    }
 
-      do {
-        let credentials = try await task.value
-        refreshTask = nil
-        consecutiveFailures = 0
-        nextRetryAt = nil
-        return credentials
-      } catch {
-        refreshTask = nil
-        await handleError(error)
-        throw error
-      }
+    guard let server, case .bearer(_, let refreshToken, _) = server.token,
+      !refreshToken.isEmpty
+    else {
+      throw Audiobookshelf.AudiobookshelfError.networkError(
+        "Refresh token is no longer valid, re-authentication required"
+      )
+    }
+
+    return try await performRefresh(for: server)
+  }
+
+  private func performRefresh(for server: Server) async throws -> Credentials {
+    let task = Task<Credentials, Error> { @MainActor [server] in
+      try await Audiobookshelf.shared.authentication.refreshToken(for: server)
+    }
+
+    refreshTask = task
+
+    do {
+      let credentials = try await task.value
+      refreshTask = nil
+      consecutiveFailures = 0
+      nextRetryAt = nil
+      return credentials
+    } catch {
+      refreshTask = nil
+      await handleError(error)
+      throw error
     }
   }
 
--- a/API/Sources/API/Models/Server.swift
+++ b/API/Sources/API/Models/Server.swift
@@ -42,6 +42,23 @@ public final class Server: @unchecked Sendable {
     }
   }
 
+  /// Whether a 401 response can potentially be recovered by refreshing the
+  /// access token. Only bearer tokens with a non-empty refresh token are
+  /// refreshable; API keys and legacy tokens are not.
+  public var canAttemptRefresh: Bool {
+    if case .bearer(_, let refreshToken, _) = token {
+      return !refreshToken.isEmpty
+    }
+    return false
+  }
+
+  /// Forces a refresh of the access token using the stored refresh token,
+  /// bypassing the normal expiry check and connection backoff. Used to recover
+  /// from a 401 on a request that should have been authenticated.
+  public func forceTokenRefresh() async throws {
+    _ = try await credentialsActor.forceRefresh()
+  }
+
   public let storage: UserDefaults
 
   @ObservationIgnored @Stored("username") public var username: String? = nil
--- a/API/Sources/API/NetworkService.swift
+++ b/API/Sources/API/NetworkService.swift
@@ -174,7 +174,10 @@ final class NetworkService {
     }
   }
 
-  private func performRequest<T: Decodable>(_ request: NetworkRequest<T>) async throws -> NetworkResponse<T> {
+  private func performRequest<T: Decodable>(
+    _ request: NetworkRequest<T>,
+    allowAuthRetry: Bool = true
+  ) async throws -> NetworkResponse<T> {
     let urlRequest = try await buildURLRequest(from: request)
 
     AppLogger.network.info(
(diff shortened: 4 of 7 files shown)


jeremygrenier:
I'm a bit confused by the changes.

I agree that we can be more resilient and only set `.authenticationError` when the server returns a 401 or 403 during token refresh, but I don't see why a server being down or unreachable would return a 401.

Also, the change feels much more complex than what it is trying to accomplish. The refresh token logic should only attempt a refresh when the token is expired. Using the current token would provide an `Authorization` header but with an expired token which should be rejected in the same way.

I'll adjust the 401 handling so that `.authenticationError` is only set during refresh operations in the next beta. Thank you for the suggestion.

jeremygrenier:
https://github.com/AudioBooth/AudioBooth/commit/a9b9044d4d4300d3741f1604ac637700dd1eeb57

spawvn:
Thanks for the quick turnaround — and fair points. Agreed the "send the stored token as a fallback" change was redundant (an expired token gets rejected just like a missing one), and [`a9b9044`](https://github.com/AudioBooth/AudioBooth/commit/a9b9044d4d4300d3741f1604ac637700dd1eeb57) is a cleaner way to get there. Glad `canAttemptRefresh` and the foreground re-probe were useful. Closing this in favor of that commit. 🙂

One edge case worth flagging for whenever you're next in this area: `a9b9044` downgrades a refreshable 401 to `.connectionError` but doesn't itself trigger a refresh — it relies on the proactive (expiry-based) refresh in `freshCredentials` to recover on a later request. That nicely covers the common case (a clock-expired access token). But if the server rejects a token the client still considers valid — server-side session revocation, a `TOKEN_SECRET` rotation, or clock skew — the proactive refresh won't fire (the token isn't "expired" by the client's clock), so requests keep getting 401s and the connection can sit in `.connectionError` indefinitely: never refreshing, never escalating to a re-login.

If you think that's worth handling, I'd be happy to open a small, focused follow-up: on a refreshable 401, force a single refresh + retry, and only fall back to `.authenticationError` if that refresh is itself rejected. Much smaller than this PR. Totally fine to leave it too — thanks again!

spawvn:
Oh, and I forgot to mention - I love the app, thank you for making it!
It's almost perfect for my needs. I was so excited to have found features I did not even know I wanted, like the automatic sleep timer. Now my son can safely forget to set the timer before going to bed, and I don't have to remember to tell him the next day that he need to rewind his book because he slept through an hour of it 😄

PR #289, opened 2026-06-24
Fix/podcast playback issues

Creationsss:
- fix auto play after playing next using old audio but new podcast name
  - make play next based on the users sort order
  - fix not updating progress unless reopening page, also fixed removing based on filters (after finishing)
  - fix libraries in dropdown not showing on initial login (had to restart app)
  - fix having to spam press play when reconnecting after session timeout (sometimes spammed reconnecting banner)
  - remove hardcoded dev team in ``project.pbxproj`` use ``Local.xcconfig``
  - fixed weird loading of podcasts list when it had it cached


jeremygrenier (approved):
Thank you for your contribution. I do not personally use ABS to listen to podcast so any help is more than welcome!

PR #292, opened 2026-06-24
Fix Smart Rewind stopping playback when resuming at a chapter start

### LINKED ISSUE #291: Smart Rewind stops playback when resuming at the start of a chapter
## Beta Check

Observed in normal use and confirmed by source analysis against current `main` — the relevant code path (`applySmartRewind`) is unchanged on `main`.

- [x] Confirmed present on current `main`.

## Description

With Smart Rewind enabled and "Don't rewind past the start of the current chapter" on (the default), pressing play to resume **at the very start of a chapter** does a negligible rewind and then **stops playback**. You have to press play a second time to actually resume.

## Steps to Reproduce

1. Settings → Player → Skip & Smart Rewind: enable Auto-rewind on Resume with the chapter barrier on (default), and a pause threshold you'll exceed.
2. Pause right at the start of a chapter.
3. Wait longer than the pause threshold.
4. Press play (e.g. from the lock screen).
5. The rewind clamps to the chapter start (barely moves), and playback stops instead of resuming.

## Expected Behavior

Playback resumes normally.

## Actual Behavior

The rewind is negligible (it can't cross into the previous chapter), and playback stops. A second press of play is needed to resume.

The play path applies Smart Rewind before resuming (`onPlayTapped` → `applySmartRewind` → `player.resume()`). At a chapter start the barrier clamps the rewind target back to the chapter start, so it moves nowhere — but `applySmartRewind` still mutates `mediaProgress.currentTime` and issues a `player.seek(...)`. That:

- wakes the `observeMediaProgress` observer, which schedules `if !isPlaying { player.seek(to: currentTime) }` on the next runloop tick, while
- `resume()` issues its own seek + `player.play()`.

`player.play()` doesn't flip the model's `isPlaying` synchronously (that happens on the async `.stateChanged(.playing)` event), so the deferred observer seek fires while `isPlaying` is still `false` and seeks again into the play-startup window — leaving playback stopped. On the second press the pause is short, so `applySmartRewind` returns early (interval 0) without seeking and resume succeeds — which is why two presses "work."

Mid-chapter resumes are unaffected (the rewind genuinely moves and playback survives), so this is specific to the no-movement / chapter-boundary case.

A deeper, related fragility: `observeMediaProgress` keys off the lagging `isPlaying` flag, so any `currentTime` change at play-startup (not just Smart Rewind) could inject a stray seek. The incoming fix only removes the user-visible trigger (the no-op rewind); hardening the observer would be a separate consideration.

## Device Information

- **Platform:** iPhone (the logic lives in the shared player layer, so it isn't device-specific)
- **AudioBooth Version:** current `main` / latest release
- **Audiobookshelf Server Version:** N/A (client-side playback logic)

## Additional Context

A fix PR follows: skip the rewind when it doesn't meaningfully move the playhead.

spawvn:
Fixes #291.

## Summary

Resuming playback at the **start of a chapter** (with "Don't rewind past the start of the current chapter" enabled) triggers Smart Rewind but then **stops playback** — you have to press play a second time to actually resume.

## Steps to reproduce

1. Enable Smart Rewind with the chapter barrier on (default) and a pause threshold you'll exceed.
2. Pause right at the start of a chapter.
3. After a pause longer than the threshold, press play (e.g. from the lock screen).
4. The rewind clamps to the chapter start (so it barely moves), and playback stops instead of resuming.

## Root cause

The play path applies Smart Rewind before resuming (`onPlayTapped` → `applySmartRewind` → `player.resume()`). At a chapter start the barrier clamps `rewindTarget` back to the chapter start, so `newTime == currentTime` — the rewind moves nowhere. But it still runs `mediaProgress.currentTime = newTime` and `player.seek(to:)`, which:

- wakes `observeMediaProgress`, which schedules `if !isPlaying { player.seek(to: currentTime) }` on the next runloop tick, while
- `resume()` issues its own `seek` + `player.play()`.

`player.play()` doesn't flip the model's `isPlaying` synchronously (that happens later, on the async `.stateChanged(.playing)` event), so when the deferred observer seek fires, `isPlaying` is still `false` and it seeks again right into the play-startup window — leaving playback stopped. On the second press the pause is short, so `applySmartRewind` computes interval 0 and returns early without seeking, and resume succeeds — which is why pressing play twice "works."

Mid-chapter resumes are unaffected (the rewind genuinely moves and playback survives the seek), so this is specific to the no-movement / chapter-boundary case.

## Fix

In `applySmartRewind`, skip when the computed rewind doesn't meaningfully move the playhead (`currentTime - newTime <= 0.5s`). With nothing to rewind, `resume()` runs cleanly and playback starts — identical to the working second-press behavior. Mid-chapter rewinds (well above the threshold) are untouched.

## Note (kept out of scope)

The deeper fragility is that `observeMediaProgress` keys off the lagging `isPlaying` flag, so it can inject a seek during play-startup. This PR intentionally only removes the trigger (the no-op rewind), which is the user-visible bug; hardening that observer against any concurrent `currentTime` change at play time would be a larger, separate change and has no other known user-facing failure today.

## Testing

Compile-verified (`xcodebuild`, iOS Simulator, `BUILD SUCCEEDED`). The root cause was derived from the source plus the reporter's first-press-fails / second-press-works observation, and the reporter confirmed mid-chapter resumes are unaffected. **Not** runtime-tested in a simulator.


jeremygrenier:
Sorry I didn't realized you already had a PR and fixed it already in https://github.com/AudioBooth/AudioBooth/commit/ad7a1f36fcf5dd6a308021a111b5d6d2d771c0ec.

spawvn:
> Sorry I didn't realized you already had a PR and fixed it already in [ad7a1f3](https://github.com/AudioBooth/AudioBooth/commit/ad7a1f36fcf5dd6a308021a111b5d6d2d771c0ec).

No problem, as long as it's fixed!

Issue #302, opened 2026-06-29
[bug/feature] Shared connection import should validate reachability and auth before saving


jeeftor:
## Summary

Shared connection import can fail without actionable diagnostics. When a user opens an `audiobooth://connection/...` link or scans a QR code, the app does not clearly distinguish whether the shared URL, server access headers such as Cloudflare Access headers, Audiobookshelf credentials, or the server itself failed.

This is especially confusing for users sharing a self-hosted Audiobookshelf server that is reachable locally without headers but requires headers when accessed remotely.

## Current behavior

- Credential-bearing shared links are decoded and immediately persisted/activated via `DeepLinkManager.handleConnectionDeepLink` -> `AuthenticationService.restoreConnection`.
- The app does not validate `/status` with the shared URL + headers before saving.
- The app does not validate `/api/authorize` before saving when Audiobookshelf credentials are included.
- Shared links without Audiobookshelf credentials prefill a login flow, but import still does not automatically test server reachability with the shared headers.
- Failures later appear as generic connection/auth problems, detached from the import action.

## Expected behavior

The receiver should get a visible import/test flow that validates the shared connection before saving or switching:

1. Decode the shared connection payload.
2. Test server reachability using the shared server URL and headers, e.g. `/status`.
3. If Audiobookshelf credentials are included, test authenticated access, e.g. `/api/authorize`.
4. Save/switch only after validation succeeds, or continue to login only after reachability succeeds.
5. Show actionable failure messages, such as:
   - Invalid shared connection link
   - Could not reach server with the shared URL and headers
   - Server reachable, but Audiobookshelf credentials failed
   - Server reachable, please sign in

jeeftor:
Opened draft PR #303 with a scoped fix for the import path.

One nuance from implementation: API-key credentials can be validated during import with `/api/authorize`, but refresh-token bearer credentials should not be proactively refreshed just to validate the link because that can rotate/consume the sender's shared refresh token. The PR therefore:

- validates shared URL + headers with `/status` before saving or opening sign-in
- validates API-key credentials with `/api/authorize` before saving
- imports refresh-token credentials only after reachability succeeds, then lets the existing auth refresh path verify them on first authenticated use
- preserves the active URL choice so a remote recipient can receive the Cloudflare-accessible URL instead of only the primary/local URL

PR: https://github.com/AudioBooth/AudioBooth/pull/303

PR #303, opened 2026-06-29
Validate shared connection imports before saving

jeeftor:
## Summary

Refs #302.

This adds validation to shared connection import so QR/deep-link receivers get immediate feedback instead of silently saving a connection that may not work.

Changes:
- Test the shared URL and custom headers against `/status` before importing or opening the sign-in flow.
- Validate shared API-key credentials against `/api/authorize` before saving.
- Preserve/export the currently active URL choice so remote recipients can receive the Cloudflare-accessible URL instead of only the primary/local URL.
- Keep custom headers independent from Audiobookshelf credentials; headers are server-access metadata and are still included in shared connection payloads.
- For refresh-token bearer links, avoid calling refresh during import because that can rotate/consume the token. The app imports after reachability succeeds and verifies Audiobookshelf credentials on first authenticated use.
- Move the shared connection payload model into the API package and add tests for the payload contract used by both QR codes and deep links.

This PR was made with the help of OpenAI Codex.

## Coverage

| Transport | Without Audiobookshelf credentials | With Audiobookshelf credentials |
| --- | --- | --- |
| Deep link | Decodes the shared connection, tests `/status` with URL + headers, then opens sign-in only if reachable. | Tests `/status` with URL + headers. API-key payloads validate `/api/authorize` before saving. Refresh-token payloads save only after reachability succeeds and are verified by the existing refresh path on first authenticated use. |
| QR code | Covered by the same payload because the QR image encodes the same `audiobooth://connection/...` URL. | Covered by the same payload because scanning the QR enters the same deep-link import path. |

## Covered by automated tests

The new `APITests` target covers the shared connection payload behavior:

- Headers are exported even when Audiobookshelf credentials are omitted.
- Tokens are omitted when `includeToken` is false.
- The active alternative URL is exported when selected.
- API-key credentials are included when credential sharing is enabled.
- Bearer credential sharing exports the refresh token, not the access token.
- Legacy credentials cannot be exported with credentials included.
- Older payloads without alternative URL fields still decode.

## Not covered / intentional limitations

- Refresh-token credentials are not proactively validated during import because validation would require refreshing, which can rotate or consume the shared token.
- The PR does not add live network tests against a real Audiobookshelf server or Cloudflare Access proxy.
- If the sender is currently using the local URL, that local URL is what gets shared. The PR preserves the active URL choice but does not add a UI prompt to choose local vs remote at share time.
- Re-import/deduplication behavior is unchanged.

## Verification

| Check | Result | Notes |
| --- | --- | --- |
| `xcrun swift-format format --in-place --parallel API/Sources/API/Services/AuthenticationService.swift AudioBooth/AudioBooth/Services/DeepLinkManager.swift API/Sources/API/Models/ExportConnection.swift API/Tests/APITests/ExportConnectionTests.swift API/Package.swift` | Pass | Formatted touched Swift/package files. |
| `xcodebuild -project AudioBooth/AudioBooth.xcodeproj -scheme API -destination 'generic/platform=iOS Simulator' -configuration Debug build` | Pass | API scheme builds successfully from the app project. |
| `xcodebuild -scheme API -destination 'generic/platform=iOS Simulator' -configuration Debug build-for-testing` from `API/` | Pass | Builds the `APITests` bundle successfully. |
| `xcodebuild -scheme API -destination 'platform=iOS Simulator,name=iPhone 17,OS=26.5' -configuration Debug test` from `API/` | Blocked | Local CoreSimulator is out of date, so concrete simulator devices are unavailable. |
| `xcodebuild -project AudioBooth/AudioBooth.xcodeproj -scheme AudioBooth -destination 'generic/platform=iOS Simulator' -configuration Debug CODE_SIGNING_ALLOWED=NO build` | Blocked | Local Xcode reports that the scheme embeds an Apple Watch app and watchOS 26.5 is not installed. |
| `xcodebuild -quiet -project AudioBooth/AudioBooth.xcodeproj -target AudioBooth -sdk iphonesimulator -configuration Debug CODE_SIGNING_ALLOWED=NO build` | Blocked | Fails before app compilation in `ReadiumShared` with missing generated `Minizip.modulemap`. |

### DIFF
--- a/API/Sources/API/Services/AuthenticationService.swift
+++ b/API/Sources/API/Services/AuthenticationService.swift
@@ -448,6 +448,42 @@ public final class AuthenticationService: ObservableObject {
     return connectionID
   }
 
+  public func loginWithJWT(
+    serverURL: URL,
+    token: String,
+    customHeaders: [String: String] = [:],
+    alias: String? = nil
+  ) async throws -> String {
+    switch JWT(token)?.type {
+    case .api:
+      let connectionID = try await loginWithAPIKey(
+        serverURL: serverURL.absoluteString,
+        apiKey: token,
+        customHeaders: customHeaders
+      )
+      updateAlias(connectionID, alias: alias)
+      try switchToServer(connectionID)
+      return connectionID
+
+    case .refresh:
+      let connection = Connection(
+        serverURL: serverURL,
+        token: .bearer(accessToken: "", refreshToken: token, expiresAt: 0),
+        customHeaders: customHeaders,
+        alias: alias
+      )
+      let server = Server(connection: connection)
+
+      _ = try await refreshToken(for: server)
+
+      restoreConnection(Connection(server))
+      return connection.id
+
+    case .access, .unknown, nil:
+      throw Audiobookshelf.AudiobookshelfError.loginFailed("Unsupported token type")
+    }
+  }
+
   func refreshToken(for server: Server) async throws -> Credentials {
     if case .apiKey = server.token {
       return server.token
--- a/AudioBooth/AudioBooth/Services/DeepLinkManager.swift
+++ b/AudioBooth/AudioBooth/Services/DeepLinkManager.swift
@@ -52,22 +52,38 @@ class DeepLinkManager: ObservableObject {
       return
     }
 
-    if let token = exportConnection.token {
-      let credentials: Credentials
-      if JWT(token)?.type == .api {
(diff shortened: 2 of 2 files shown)


jeremygrenier:
The PR was over complicated for what it was trying to accomplished. I fixed it and will send it to the next beta. Thank you for the contribution.

jeremygrenier (approved):

PR #307, opened 2026-07-04
Focus search field when opening the search tab

octopotato:
Opening the search tab dropped you on the results view with the field unfocused, so searching was always tab, tap the field, then type. I wanted to kill that extra tap.

I focus the field on appear when there's no query yet, so switching to the tab puts the cursor straight in the search box. I guard on empty text so coming back to the tab with an existing search doesn't steal focus or pop the keyboard.

https://github.com/user-attachments/assets/665b3f6f-81aa-4a54-be5e-8a99194aa8d6

### DIFF
--- a/AudioBooth/AudioBooth/Screens/Search/SearchView.swift
+++ b/AudioBooth/AudioBooth/Screens/Search/SearchView.swift
@@ -2,14 +2,22 @@ import API
 import Combine
 import SwiftUI
 
+@available(iOS 26.0, *)
 struct SearchPage: View {
   @StateObject var model: SearchView.Model
+  @FocusState private var searchFieldFocused: Bool
 
   var body: some View {
     NavigationStack {
       SearchView(model: model)
         .searchable(text: $model.searchText)
+        .searchFocused($searchFieldFocused)
         .navigationDestination(for: NavigationDestination.self) { $0.resolvedView }
+        .onAppear {
+          if model.searchText.isEmpty {
+            searchFieldFocused = true
+          }
+        }
     }
   }
 }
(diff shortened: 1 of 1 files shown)


AudioBooth/AudioBooth/Screens/Search/SearchView.swift
```diff
@@ -2,14 +2,22 @@ import API
...
+@available(iOS 26.0, *)
 struct SearchPage: View {
   @StateObject var model: SearchView.Model
+  @FocusState private var searchFieldFocused: Bool
 
   var body: some View {
     NavigationStack {
       SearchView(model: model)
         .searchable(text: $model.searchText)
+        .searchFocused($searchFieldFocused)
```
jeremygrenier: I wasn't even aware of ` .searchFocused()`, I tried to use `.searchable(text:, isPresented:)` but it was buggy on iOS 26.0.

AudioBooth/AudioBooth/Screens/Search/SearchView.swift
```diff
@@ -2,14 +2,22 @@ import API
 import Combine
 import SwiftUI
 
+@available(iOS 26.0, *)
 struct SearchPage: View {
   @StateObject var model: SearchView.Model
+  @FocusState private var searchFieldFocused: Bool
```
jeremygrenier: We are already on the `SearchPage` so `fieldFocused` or even `focused` seems cleaner to me.

jeremygrenier (approved):
Thank you.

AudioBooth/AudioBooth/Screens/Search/SearchView.swift
octopotato: Fair, I used the same naming as the ebook variant 😛 
https://github.com/octopotato/AudioBooth/blob/22908fd0e1b0a7b74aabdd25399712a162ee2d9f/AudioBooth/AudioBooth/Screens/EbookReader/EbookSearch/EbookSearchView.swift#L8

PR #308, opened 2026-07-04
Add continue-listening fallback to auto-play next

octopotato:
Auto-play next went straight from next-in-series to the next downloaded book, so finishing a standalone book (or last book in a series) jumped to whatever happened to be next in downloads instead of what I was actually working through which is a bit jarring. I wanted it to reach for continue-listening before falling back to the download order since I usually listen to multiple books in parallel depending on my mood, so the next most recent listen is what I would expect to be picked up. The auto play next behavior becomes:
1. next in series
2. next in continue listening
3. next in downloads

### DIFF
--- a/AudioBooth/AudioBooth/Services/SmartContinueResolver.swift
+++ b/AudioBooth/AudioBooth/Services/SmartContinueResolver.swift
@@ -27,6 +27,9 @@ struct SmartContinueResolver {
       if let next = await resolveNextBookInSeries(currentBookID: currentItemID) {
         return next
       }
+      if let next = resolveNextBookInContinueListening(currentBookID: currentItemID) {
+        return next
+      }
       return resolveNextOfflineBook(currentBookID: currentItemID)
     }
   }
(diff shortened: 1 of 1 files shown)


jeremygrenier:
I think having `continue listening` part of the smart continue make sense however for me downloads take the priority over the `continue listening`.

I personally download the next book I want to listen to before I complete my current one but my continue listening section is populated with DNF books.

octopotato:
Interesting.. I think my problem is twofold:
1. either I haven't figure it out how it works correctly or I'm using it wrong.. I can't make heads or tails of how the download section is ordered. It _feels_ like downloads ordering is.. random? haphazard? maybe it downloads immediately after the most recently listened to downloaded book? Quickly looking at the code it seems like:
    - ordered by `createdAt` asc order (oldest to most recent)
    - if order was manually changed at one point, then new downloads will go in 2nd place and after since first place has `displayOrder = 0` and the tie breaker is `createdAt` which messes up my previous ordering ex:
        ```
        #1  original 1     (displayOrder 0)
        #2  new book A     (displayOrder 0, createdAt = oldest of the 3)
        #3  new book B     (displayOrder 0)
        #4  new book C     (displayOrder 0, newest)
        #5  original 2     (displayOrder 1)
        #6  original 3     (displayOrder 2)
        #7  original 4     (displayOrder 3)
        ...
        ```
2. downloaded at section isn't as visible/front and center as continue listening, so I usually have continue listening tidy since it's always in my face (aka on my homepage) where as downloaded is more of a "in the background"

I think I'd be more okay with download order if it didn't screw up my manual ordering. Maybe all new downloads go to the bottom of the list if the downloads were previously ordered then auto continue plays the next book in line or top of the list

jeremygrenier:
That is odd. New downloads should always appear at the end of the list, unless you're collapsing by series. If that's not the case, then we likely have a bug.

You can also drag and drop items to reorder them however you'd like.

octopotato:
Yea just tested it and the new download showed up as the second book in my ordered download list which is annoying without collapsing by series (although I just found out collapse by series is a thing for downloads section 😆 so that'll be useful)

PR #318, opened 2026-07-09
fix: crash when opening Custom Headers while adding first server on iOS 18

### LINKED ISSUE #312: Crash when trying to add a custom header
## Beta Check
Before opening this issue, please test if the bug is still present in the latest beta version available on [TestFlight](https://testflight.apple.com/join/cAG6dVeN). Investigating issues that are already fixed takes time away from making the app better.

- [x] I have tested on the latest beta and the issue is still present.

## Description
The app becomes unresponsive and crashes when trying to add the first server and tapping the “Custom Headers” button. This happened after having the server signed in and working, then signing out to re-authenticate using a custom header (after having made changes to the server that requires a custom header).

The issue is either iOS version specific OR device specific. See details below.

## Steps to Reproduce
1. Do a fresh install on a device running iOS 18.3.2
2. Tap “Add Server”
3. Tap “Custom Headers”
4. RESULT - the app freezes and crashes.

## Expected Behavior
The app doesn’t crash.

## Actual Behavior
The app crashes.

## Screenshots
NA

## Device Information
- **Platform:** iPhone
- **Device Model:** iPhone 16 Pro
- **OS Version:** iOS 18.3.2
- **AudioBooth Version:** 1.11
- **Audiobookshelf Server Version:** NA

## Additional Context

This issue does NOT occur on my iPhone 15 pro running iOS 26.5.

I tried uninstalling and reinstalling the app, restarting the device and terminating/relaunching the app. After all the steps listed, it behaves the same.

mvanhorn:
## Summary

Opening Custom Headers while adding the first server crashes on iOS 18. `CustomHeadersView` declared the shared observable server model as `@StateObject`, but the model is `@Observable` (not an `ObservableObject`), and on iOS 18 SwiftUI traps when a not-yet-persisted model is wrapped that way from the add-server sheet. Declaring it `@Bindable` matches how the model is passed everywhere else and stops the crash.

Fixes #312

### DIFF
--- a/AudioBooth/AudioBooth/Screens/Servers/Server/CustomHeaders/CustomHeadersView.swift
+++ b/AudioBooth/AudioBooth/Screens/Servers/Server/CustomHeaders/CustomHeadersView.swift
@@ -4,7 +4,7 @@ import SwiftUI
 struct CustomHeadersView: View {
   @Environment(\.appTheme) var theme
 
-  @StateObject var model: Model
+  @Bindable var model: Model
 
   var body: some View {
     List {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Tip jar for sponsors - Tip jar section now collapses when you have an active sponsor subscription
 
 ### Fixed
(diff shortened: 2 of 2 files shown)


jeremygrenier:
The model is an `ObservableObject` but you are right it should be a `@ObservedObject` and not a `@StateObject`. I will fix it in the next beta.

PR #321, opened 2026-07-11
Sync Playing Next queue across devices via iCloud

### LINKED ISSUE #320: [Feature Request] Sync Playing Next queue across devices via iCloud
## Feature request

Currently the "Playing Next" queue (`PlayerManager.queue`) is stored only in local `UserDefaults` on each device. If I add episodes to the queue on my iPhone, they don't show up on my iPad, and vice versa.

### Desired behavior

Adding, removing, or reordering items in the Playing Next queue on one device should be reflected on the user's other devices signed into the same iCloud account, when connected to the same Audiobookshelf server.

### Proposed approach

The app already has an iCloud sync mechanism for preferences (`UserPreferences+Cloud.swift`) built on `NSUbiquitousKeyValueStore`, gated behind the existing "iCloud Sync" toggle in Advanced Preferences. The queue can piggyback on the same mechanism:

- Mirror `PlayerManager.queue` through `NSUbiquitousKeyValueStore`, scoped per server (using the server's base URL, since the per-connection `Server.id` is a locally generated UUID and differs across devices for the same server).
- Guard against echo loops when applying a remote change.
- Avoid clobbering another device's queue on first launch / server switch.

### Scope

Sync the whole queue (books and podcast episodes both use the same `QueueItem`/`PlayerManager.queue`), not just podcast episodes.

I have a working implementation and will open a PR against this.

robnightingale:
Closes #320

## Summary
- Mirror the "Playing Next" queue (`PlayerManager.queue`) through the existing `NSUbiquitousKeyValueStore` iCloud mechanism, gated behind the existing iCloud Sync preference, so adding/removing/reordering items on one device shows up on the user's other devices.
- Scope the synced queue per server, keyed by the server's normalized base URL (not `Server.id`, which is a locally generated UUID that differs per device for the same server — this was caught during real cross-device testing and fixed in the second commit).
- Guard against echo loops when applying a remote change, and avoid an empty queue on a fresh launch/install clobbering another device's queue before the initial pull completes.
- Toggling iCloud Sync off purges the synced queue from the cloud.

## Test plan
- [x] Builds clean (`xcodebuild build`, iOS Simulator destination)
- [x] Verified end-to-end between iPhone 17 Pro and iPad mini simulators on the same server/Apple ID, forcing delivery with `xcrun simctl icloud_sync`: adding items on one device correctly appeared on the other after the fix
- [ ] Verify on two real devices (simulator iCloud KVS delivery is not fully representative of real-device behavior)


jeremygrenier:
Hello,

The current iCloud implementation is only used to sync app preferences. It doesn't sync anything tied to a specific server connection or library.

For now, I'd like to avoid syncing server-specific data through iCloud. I may add an option to sync connections across devices in the future, but that's not something the app currently supports.

If you're looking for a similar workflow, you could use a pinned playlist on both devices and treat it as your queue.

robnightingale:
Thank you for your response. That's fair enough, I'll try that approach, and close the PR.  Thanks again.

PR #323, opened 2026-07-14
Keep long-press menu actions near the cover

octopotato:
Long-pressing a book or episode always pinned the primary actions (Play, Download) to the top of the menu, so when the menu opened above the cover they became the farthest thing to reach instead of the closest. Annoying on taller screens where cards near the bottom always open the menu upward.

I switched the long-press menus to priority ordering so the OS keeps the primary actions closest to your touch point and flips them when the menu opens upward.

| Before | After |
|--------|--------|
| Screenshot 2026-07-13 at 4 54 58 PM | Screenshot 2026-07-13 at 4 59 40 PM |


jeremygrenier (approved):
Looks good to me. I wasn't aware of `.menuOrder(.priority)`.

PR #330, opened 2026-07-21
Fix background downloads for audiobooks with many files

cbzehner:
First, thank you so much for building this! I am a very happy user of AudioBooth. 👋 In particular, the ability to see my play/pause History ↺ is one of my favorite features, I wish every audio app featured this sort of granular seek system!

Second, I'm not an expert on iOS development or Swift, so please give feedback and let me know what changes you'd request before deeming this ready to merge, or if you'd prefer that I close the PR out. I've established that this resolves the issue I was facing with long download times locally on my phone when I install the app with this patch and re-download the content.

## Problem

AudioBooth starts each audiobook file only after the prior file finishes.

iOS limits how often an app can start a task in a background session. A book with many files can therefore pause for several minutes between files. The download can appear stuck or can fail.

## Steps to reproduce

1. Add an audiobook that has about 100 separate audio files.
2. Open the book in AudioBooth.
3. Start the download.
4. Keep the app open or move it to the background.
5. Watch the download after the first files finish.

### Actual result

Later files can take several minutes to start. The download may appear stuck or may fail.

### Expected result

AudioBooth should submit the full download to the background session. iOS should then schedule the files without a long pause between each file.

## Change

- Create all file download tasks before any task starts.
- Let the background session schedule the full batch.
- Track each task by its task identifier.
- Combine progress from all active tasks.
- Retry each failed task with resume data when iOS provides it.
- Cancel all active tasks when the user cancels the book.
- Complete the book only after all files finish.
- Use the same download path for ebooks and podcast episodes.

## Test results

- Downloaded a 1.39 GB book with 106 audio files and one ebook.
- Completed a full download in 10 minutes and 37 seconds.
- Completed a second full download with 106 audio files and zero failed requests.
- Confirmed that user cancellation stops all active requests.
- Confirmed that downloads restart after a short server outage.
- Saw no app crash.
- Confirmed that the server stayed healthy during the download.
- Passed the iOS Simulator build.
- Passed the strict Swift format check.

## Scope

This PR changes `DownloadManager.swift` only. It does not change the server API or the stored book format.

## LLM usage

I used an LLM to help:

- trace the download flow;
- identify the background task scheduling fault;
- draft and simplify the code change;
- review the diff for excess complexity;
- prepare test commands; and
- monitor logs during device tests.

A person directed the work and approved the approach. The PR author reviewed the final diff. The change was built and tested on a physical iPhone against an Audiobookshelf server.

### DIFF
--- a/AudioBooth/AudioBooth/Services/DownloadManager.swift
+++ b/AudioBooth/AudioBooth/Services/DownloadManager.swift
@@ -299,11 +299,22 @@ private final class DownloadOperation: Operation, @unchecked Sendable {
 
   private let maxRetryAttempts = 3
 
-  private var currentTrack: URLSessionDownloadTask?
-  private var continuation: CheckedContinuation<Void, Error>?
+  private struct FileDownload {
+    let request: URLRequest
+    let expectedSize: Int64
+    let destination: URL
+  }
+
+  private struct ActiveDownload {
+    let file: FileDownload
+    let attempt: Int
+    let task: URLSessionDownloadTask
+  }
+
+  private var activeDownloads: [Int: ActiveDownload] = [:]
+  private var taskBytesWritten: [Int: Int64] = [:]
+  private var batchContinuation: CheckedContinuation<Void, Error>?
   private let continuationLock = NSLock()
-  private var trackDestination: URL?
-  private var lastResumeData: Data?
 
   private lazy var downloadSession: URLSession = {
     let config = URLSessionConfiguration.background(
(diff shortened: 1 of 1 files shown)


jeremygrenier:
Hello, thank you for the contribution! It was a great start. I was rebasing your PR when I started digging more into revamping the download logic as a whole. I then spent a few days working on it before going on vacation.

Sorry for the delay, but I took over your PR and added several improvements. I’ll close this PR because my changes will be merged into `main` soon.

Thank you again for your contribution!

PR #337, opened 2026-07-22
Keep the currently playing download when marking as finished

### LINKED ISSUE #336: Remove Download on Completion deletes the currently playing book's files, breaking immediate re-listen
## Beta Check

Confirmed by source analysis against current `main` (`dacfde1`) — the code path is unchanged there.

- [x] Confirmed present on current `main`.

## Description

With **Settings → Storage → "Remove Download on Completion" enabled**, finishing a downloaded book deletes its files **at the moment of completion, while the book is still loaded in the player**. If you then try to re-listen (e.g. jump back a couple of chapters right after the book ends), the player is left pointing at files that no longer exist and gets stuck doing nothing — no error, no recovery. Only re-opening the book from scratch recovers.

## Steps to Reproduce

1. Enable "Remove Download on Completion" (Settings → Storage).
2. Download a book and play it to the end (Smart Continue off makes it easy to observe — the finished book stays open).
3. Immediately seek back (e.g. via the chapter picker) and press play.
4. Playback never starts; the player is unresponsive with no error shown.

At completion, `recordBookCompletionIfNeeded` → `LocalBook.markAsFinished()` (`BookActionable.swift`), which — when the preference is on and the book is downloaded — calls `removeDownload()` → `DownloadManager.deleteDownload(for:)`. That deletes the audio files **and** the `LocalBook` record **unconditionally**, even though this book is still the currently loaded player:

- The `PlayerManager.shared.current?.id != bookID` guard inside `BookActionable.removeDownload()` only wraps a *second, redundant* record deletion — `deleteDownload` has already removed both the files and the record, so the guard protects nothing.
- The player's `PlaybackSession.url(for:)` returns `track.localPath` without checking the file still exists, so any subsequent seek/resume rebuilds `AVPlayerItem`s from deleted file URLs.
- The resulting failure dead-ends silently: `handleStreamFailure` hits `guard item?.isDownloaded != true else { return }` ("Book is downloaded, cannot recover from stream failure") — a debug log, no toast, no recovery. The session-recreation path in `onPlayTapped` likewise skips reloading when the player uses local URLs.

Notably, the **batch** cleanup for this same preference — `DownloadManager.removeCompleted()`, run on every Home remote fetch — already **skips the currently playing item** (`bookID != currentPlayingID`). The completion path just bypasses that rule.

## Fix (PR follows)

Apply the same guard in the completion path: don't delete the download when the finished book is still the currently loaded player. Nothing is lost — `removeCompleted()` already sweeps finished downloads on the next Home fetch, once the book is no longer loaded. Deletion is deferred, not skipped.

## Possible follow-up hardening (out of scope of the PR)

- `handleStreamFailure` could detect "local book whose files are actually gone" and fall back to streaming (or at least surface an error toast) instead of returning silently.
- `PlaybackSession.url(for:)` could verify file existence before preferring `localPath`.

spawvn:
Fixes #336.

## Summary

With "Remove Download on Completion" enabled, finishing a downloaded book deletes its files (and `LocalBook` record) **at the moment of completion, while the book is still loaded in the player**. Trying to re-listen right after the book ends (e.g. jumping back a couple of chapters) leaves the player rebuilding `AVPlayerItem`s from deleted file URLs — it silently gets stuck: no playback, no error, no recovery.

## Root cause

`markAsFinished()` (`BookActionable.swift`) calls `removeDownload()` → `DownloadManager.deleteDownload(for:)`, which removes the files and the `LocalBook` record unconditionally. The existing `PlayerManager.shared.current?.id != bookID` check inside `removeDownload()` only guards a second, redundant record deletion, so it doesn't protect the currently playing book.

Meanwhile the batch cleanup for the same preference — `DownloadManager.removeCompleted()`, run on every Home remote fetch — already skips the currently playing item. The completion path just bypassed that rule.

## Fix

Apply the same guard in the completion path: skip the immediate deletion when the finished book is still the current player.

Nothing is left behind — `removeCompleted()` deletes the finished download on the next Home fetch, once the book is no longer loaded. Deletion is deferred, not lost. Books finished while *not* loaded in the player (e.g. marked finished from the context menu) are unaffected and still delete immediately.

## Testing

Compile-verified (`xcodebuild`, iOS Simulator, `BUILD SUCCEEDED`) and `swift-format` clean. The root cause and fix were derived from source analysis of a reproducible user report (finish downloaded book → seek back → player stuck); **not** runtime-tested on a device. The change mirrors the guard `removeCompleted()` already uses, so the deferred-cleanup behavior is the codebase's existing pattern.

### DIFF
--- a/AudioBooth/AudioBooth/Services/BookActionable.swift
+++ b/AudioBooth/AudioBooth/Services/BookActionable.swift
@@ -17,7 +17,13 @@ extension BookActionable {
     try await Audiobookshelf.shared.libraries.markAsFinished(bookID: bookID)
 
     if UserPreferences.shared.removeDownloadOnCompletion {
-      if DownloadManager.shared.downloadStates[bookID] == .downloaded {
+      // Deleting while this book is still loaded would leave the player pointing
+      // at files that no longer exist (e.g. re-listening right after finishing).
+      // Skip it here — removeCompleted() sweeps finished downloads on the next
+      // home fetch, once the book is no longer the current player.
+      if DownloadManager.shared.downloadStates[bookID] == .downloaded,
+        PlayerManager.shared.current?.id != bookID
+      {
         removeDownload()
       }
     }


jeremygrenier:
Thank you for the report and the PR. The code is correct but we do not need all the comment. I will fix it in the next beta 🙇

PR #347, opened 2026-08-08
Add Options for Headphone Controls and New Shortcut to Return Current Item Playing

aakar:
I've tested this on my iPhone and it seems to work well. It does not impact the Lock Screen Fast Forward. However, I have not tested this with CarPlay.

This is what the branch has:

Headphone controls: a dedicated settings screen that lets you override the next/previous headphone gestures. Options include Skip, Rewind, Next/Previous Chapter, Add Bookmark, and Nothing. This is configurable independently per gesture, with .default preserving existing behavior.
Add-bookmark action: Creates a bookmark from the controls, with a haptic, a toast, and a subtle chime that mixes over playback without pausing the book. Auto-titled with a date/time.
Tests: a new ModelsTests target covering the gesture-routing resolver and preference round-tripping.
"Get Current Item" shortcut: a new App Intent / Siri shortcut returning the currently playing item with title, author, chapter, and position. This can be ignored; I was testing some things and thought it would be good to have the Current Item returned via Shortcuts (dropped Previous Chapter's phrase for the new one


jeremygrenier:
I'm closing the PR for now, as I don't plan to add this to the codebase anytime soon. Thank you for your suggestion and contribution!

PR #350, opened 2026-08-10
Filter out authors with zero books from Authors list

### LINKED ISSUE #328: Request: Remove Authors from Library>Authors view with zero books
I believe this a bug with the Audiobookshelf library itself, but it would be good it Audiobooth could automatically hide Authors from the Library>Authors view who have zero books.

Typically these are author's where the metadata was incorrect, and the metadata in the file was fixed, the book was removed from the library, and then re-addedimporting the correct metadata. Sometimes in theis situation the old author remians in the library with zero books.

I don't know if the same issue can happen with Narrator, since my library does not have any with this issue, but it might be good to include a similar check just in case.

kornelko2:
Closes #328.

Authors with zero books (e.g. left behind after a metadata fix + re-import)
were still showing up in Library > Authors. `AuthorCard.Model` already
carries `bookCount` and the row view already hides the "N books" label when
it's 0 — the list just wasn't filtering them out. Filters `bookCount > 0`
before building the sections.

Only touched the Authors list, since the request only reported it there —
happy to add the same filter for Narrators if that turns out to need it too.

### DIFF
--- a/AudioBooth/AudioBooth/Screens/Library/Authors/AuthorsPageModel.swift
+++ b/AudioBooth/AudioBooth/Screens/Library/Authors/AuthorsPageModel.swift
@@ -60,7 +60,9 @@ final class AuthorsPageModel: AuthorsPage.Model {
         return
       }
 
-      let authorCards = response.results.map { AuthorCardModel(author: $0) }
+      let authorCards = response.results
+        .map { AuthorCardModel(author: $0) }
+        .filter { $0.bookCount > 0 }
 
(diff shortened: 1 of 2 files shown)


kornelko2:
Pushed a second commit: same fix, applied to the Narrators list too — you mentioned in the issue you weren't sure if it applied there. `NarratorCardModel` has the same `bookCount`, and the row view already hides the label at 0, so it was the same one-line filter.

jeremygrenier (approved):
LGTM! 🙏

PR #352, opened 2026-08-10
Pad series/genre/tag list content past the mini player

### LINKED ISSUE #305: Too many books in a series makes the final one impossible to select
## Beta Check
Before opening this issue, please test if the bug is still present in the latest beta version available on [TestFlight](https://testflight.apple.com/join/cAG6dVeN). Investigating issues that are already fixed takes time away from making the app better.

- [x] I have tested on the latest beta and the issue is still present.

## Description
Some ebook series have enough titles that the final one is blocked by the now playing pane and scrolling down does not rectify

## Steps to Reproduce
1. From a title's page, tap the series name to view all titles in the series
2. Scroll to the bottom, attempting to open the final title

## Expected Behavior
Final title populates above the now playing pane, making the title selectable

## Actual Behavior
The page bounces up, covering the final title with the now playing pane

## Screenshots
image

## Device Information
- iPhone 
- 14 Pro
- iOS 27 developer beta 2
- App Version 1.11 (1782969537)
- **Audiobookshelf Server Version:** 2.35.1 [Docker]

kornelko2:
Closes #305.

`LibraryPage` backs the series/genre/tag/narrator/author-library detail
lists (`NavigationDestination.resolvedView`), and its `ScrollView` had no
bottom padding of its own — it was relying entirely on
`tabViewBottomAccessory`'s automatic safe-area inset to keep content clear
of the mini player. That doesn't seem to reliably reach pushed
`NavigationStack` destinations (this shows up on iOS 26 with
`tabBarMinimizeBehavior(.onScrollDown)`), so with enough items the last row
ends up permanently behind the mini player with no way to scroll it clear.

This adds explicit bottom padding to the scroll content whenever a player is
active, as a buffer past whatever the system's own inset does or doesn't
account for.

**Caveat:** I don't have a library large enough to reproduce the original
report, so I couldn't visually confirm this on-device — this is based on
reading the layout code, not a confirmed repro. The 80pt figure is a rough
buffer, not measured against the actual mini-player height, so please treat
it as a starting point to adjust rather than a precise fix if you get a
chance to check it against a real device.

### DIFF
--- a/AudioBooth/AudioBooth/Screens/Library/Library/LibraryPage.swift
+++ b/AudioBooth/AudioBooth/Screens/Library/Library/LibraryPage.swift
@@ -5,7 +5,6 @@ import SwiftUI
 struct LibraryPage: View {
   @Environment(\.appTheme) var theme
   @ObservedObject private var preferences = UserPreferences.shared
-  @ObservedObject private var playerManager = PlayerManager.shared
 
   @ObservedObject var model: Model
 
@@ -276,10 +275,7 @@ struct LibraryPage: View {
         }
       }
       .padding(.horizontal)
-      // tabViewBottomAccessory's safe-area inset doesn't always reach pushed
-      // NavigationStack destinations reliably, so the last row can end up
-      // permanently hidden behind the mini player. Pad past it explicitly.
-      .padding(.bottom, playerManager.hasActivePlayer ? 80 : 0)
+      .padding(.bottom, 80)
       .environment(\.itemDisplayMode, preferences.libraryDisplayMode)
     }
   }
--- a/AudioBooth/AudioBooth/Screens/Library/Library/LibraryPage.swift
+++ b/AudioBooth/AudioBooth/Screens/Library/Library/LibraryPage.swift
@@ -5,6 +5,7 @@ import SwiftUI
 struct LibraryPage: View {
   @Environment(\.appTheme) var theme
   @ObservedObject private var preferences = UserPreferences.shared
+  @ObservedObject private var playerManager = PlayerManager.shared
 
   @ObservedObject var model: Model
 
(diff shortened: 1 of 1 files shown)


kornelko2:
For context on why I went with a manual padding workaround rather than trusting the system inset: `tabViewBottomAccessory` (iOS 26) has a track record of safe-area/visibility rough edges reported on Apple's Developer Forums, not just in this app:

- [Possible iOS 26 bug: tabViewBottomAccessory not affected by toolbarVisibility](https://developer.apple.com/forums/thread/795786)
- ['tabViewBottomAccessory' leaves an empty accessory area when conditionally hidden](https://developer.apple.com/forums/thread/803428)
- ['tabViewBottomAccessory' leaves an empty container when conditionally hidden](https://developer.apple.com/forums/thread/803393)
- [iOS 26 tabViewBottomAccessory doesn't work with tabViewSidebarBottomBar](https://developer.apple.com/forums/thread/800795)

None of these are an exact match for this issue (last row of a pushed `NavigationStack` list hidden behind the accessory), so I can't say for sure this app is hitting the same root cause — but it's the same general area of the API (accessory height/visibility not reliably reflected in the safe area it reports), and per a DTS engineer response in one of those threads, some of it wasn't even controllable via API until the 26.2 beta. Figured it was worth linking in case it saves time confirming this isn't app-specific.

AudioBooth/AudioBooth/Screens/Library/Library/LibraryPage.swift
```diff
@@ -275,6 +276,10 @@ struct LibraryPage: View {
         }
       }
       .padding(.horizontal)
+      // tabViewBottomAccessory's safe-area inset doesn't always reach pushed
+      // NavigationStack destinations reliably, so the last row can end up
+      // permanently hidden behind the mini player. Pad past it explicitly.
+      .padding(.bottom, playerManager.hasActivePlayer ? 80 : 0)
       .environment(\.itemDisplayMode, preferences.libraryDisplayMode)
```
jeremygrenier: To be honest I feel we can always add the 80 padding, worst case you get some extra padding at the bottom which is not a huge deal, but that avoids having the page observe the `playerManager` just for that. Also we do not need the comment, if someone wants to know why they can always use git blame and find this PR to understand the reason.

jeremygrenier (changes requested):

kornelko2:
Done — always 80, and the `playerManager` observation went with it.

Fair on the comment too. When I am explaining something to myself, or I am not sure about it, I tend to park it in the code. Every so often I sweep them out with a script — too many of them irritates me as well. So no attachment, easy to drop.

jeremygrenier (approved):

PR #358, opened 2026-08-14
added configurable behavior for sleep timer when playback is paused

fivezeroes:
Added new new options that change the default behavior for the sleep timer when media is paused. Previously, the timer continued regardless of playback state, and that behavior has been maintained as "none" and kept as the default setting. Additional settings were added that either pause the timer when playback is paused, reset the timer (good for an alternative to shaking the device), or turn the timer off.
image


jeremygrenier (approved):
Thank you for the contribution and the nice new preference!

PR #360, opened 2026-08-15
added a custom duration option for auto timer

fivezeroes:
Added a custom duration picker for the auto sleep timer
image
image
image


jeremygrenier:
Thank you for the PR. I'm wondering if we might be going a bit too far with this. I understand that some people may want a custom time, but we're already offering a lot of options, and adding a custom option increases the complexity and maintenance burden without providing much additional value.

fivezeroes:
> Thank you for the PR. I'm wondering if we might be going a bit too far with this. I understand that some people may want a custom time, but we're already offering a lot of options, and adding a custom option increases the complexity and maintenance burden without providing much additional value.

I was thinking about this too, it might make more sense to simplify the timer altogether by switching to just a custom time picker or chapter ends, and either moving it out of the player's quick settings or making auto timer follow the previous timer behavior. Its ultimately up to you, but I'm not a fan of custom times but only for the manual sleep timer.

PR #363, opened 2026-08-23
Recover playback after xHE-AAC decode failures

octopotato:
xHE-AAC audiobooks play fine until you seek, and then playback dies. Apple's USAC decoder reports `-11821` at specific positions in files that aren't fully conformant (FB22340742), so the player just paused and stayed paused until you manually seeked somewhere else.

I ended up building a probe to figure out what actually recovers it, because the obvious fix doesn't work. On a real book I found 22 failing positions out of 120 sampled, and the decoder failed identically whether the seek was frame accurate or given a second of tolerance, so seek tolerance is a red herring. Retrying at the same spot fails too, since the decoder stays bad until the queue is rebuilt. Rebuilding one second earlier per attempt cleared every position I hit, so that's what playback now does when it sees a decode failure.

Worth noting the same Apple bug was reported against ShelfPlayer in rasmuslos/ShelfPlayer#439, which is where the radar number comes from.

> [!NOTE]
> Note the player controls are only greyed out in this video to highlight what is happening, they're not actually greyed out in the final product
 
https://github.com/user-attachments/assets/f169f756-4bef-4680-8d5d-2c27227a7193

https://github.com/user-attachments/assets/0a3e770c-3394-4eb1-a3e2-5461f7385360

> [!IMPORTANT]
> This doesn't fix all issues just does a best effort since apple doesn't always send a failed decode error (i.e. on seek sometimes it plays nothing with no errors for a few seconds even before this change since apple didn't report it as a decode error until it reaches a "fixed" frame

### DIFF
--- a/AudioBooth/AudioBooth/Services/DownloadManager.swift
+++ b/AudioBooth/AudioBooth/Services/DownloadManager.swift
@@ -225,6 +225,7 @@ final class DownloadManager: NSObject, ObservableObject {
   }
 
   func resumeOutstandingRequests() {
+    guard ModelContextProvider.shared.activeServerID != nil else { return }
     guard NetworkMonitor.shared.interfaceType == .wifi else { return }
     startNextIfIdle()
   }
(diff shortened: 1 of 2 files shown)


AudioBooth/AudioBooth/Services/DownloadManager.swift
```diff
@@ -225,6 +225,7 @@ final class DownloadManager: NSObject, ObservableObject {
   }
 
   func resumeOutstandingRequests() {
+    guard ModelContextProvider.shared.activeServerID != nil else { return }
```
octopotato, line 228: had to add this to prevent app from crashes for a stale server connection on my simulator

jeremygrenier:
Thank you for the PR however I already fixed it yesterday. I will close this PR.

Issue #366, opened 2026-09-03
Feature: Page Sync — scan a page of the printed book to jump to it in the audiobook


andrespinones:
Hi Jeremy, first of all thanks for AudioBooth; it has been my daily Audiobookshelf client for a while.

I read physical books and listen to the same titles as audiobooks, and I kept losing my place when switching. Audible recently shipped "Page Sync" and Spotify "Page Match" for exactly this, so I built it for AudioBooth on top of what the app already has (Readium for the ebook text, the player, the library search). It is working on my iPhone and I would like to contribute it if you are interested.

**What it does**

From the player menu, "Page Sync" opens the camera. Point it at the page you are reading in the printed book: AudioBooth reads the page, finds it in the ebook edition from your library, then listens to the audiobook around that spot to land on the exact sentence, and offers to play from there. In my tests it takes 5–15 seconds and lands within a couple of seconds of the page start.

**How it works**

- VisionKit live text reads the page and captures on its own once it has enough words.
- Short word windows from the page are searched verbatim in the ebook text (Readium content service). Several agreeing windows locate the page.
- The ebook section is mapped to an audio chapter to get a few candidate positions.
- The audio around each candidate is transcribed on device (SpeechAnalyzer on iOS 26, SFSpeechRecognizer as fallback) and aligned with the page words until the page is heard.
- The measured drift per book is stored so later scans hit on the first try.

**Requirements and limits**

- Needs both the audiobook and the ebook in the Audiobookshelf library. They can be separate items with the same title; the app pairs them.
- Adds camera and speech recognition usage descriptions to Info.plist (App Store privacy labels would need updating).
- Streamed audio is fetched to a temp file for the alignment; single-file audiobooks over 400 MB prompt for the regular download instead.
- No new dependencies; everything runs on device.

**Proposed contribution**

1. PR: Page Sync (live camera flow, a "Page Sync" entry in the player menu shown only when the library has the ebook edition, and a one-time intro sheet). About 2,600 lines, almost all in new files under `Services/PageSync` and `Screens/PageSync`, plus a menu item in the player. Branch: https://github.com/andrespinones/AudioBooth/tree/feature/page-sync
2. Later PR, if you want it: "Read & Listen", the same machinery from the ebook reader ("Listen where I'm reading") and the player ("Read where I'm listening"), a card on Book Details grouping the three actions, and optionally an automatic "you read further in the ebook, continue from there?" offer.

**Demo video**

[...]

PR #367, opened 2026-09-04
Add Page Sync: scan a page of the printed book to jump to it in the audiobook

andrespinones:
Implements the feature proposed in #366. Draft while we settle the approach.

**Summary**

The player menu gets a "Page Sync" entry for audiobooks whose title also has an ebook in the library. It opens a live scanner; once the page is read, AudioBooth finds it in the ebook text, listens to the audiobook around the estimated spot to pinpoint it, and offers to play from there. A one-time intro sheet explains what it does and what it needs.

**Built on Read Along**

The text and speech work reuses the Read Along services as they are, no changes to them:

- `BookTextIndex` + `TranscriptAligner` locate the scanned page in the book (the page words are the query) and provide the passage shown in the result.
- `NarrationTranscriber` transcribes the audio around the candidate position; the transcript is aligned with the book with `TranscriptAligner`. The first aligned chunk tells how far the narration is from the text estimate, so the search jumps straight to the corrected time instead of scanning chapter by chapter.
- `ReadAlongAvailability` prepares the speech model.

Like Read Along, exact positioning needs iOS 26 and the audiobook downloaded. Otherwise the sheet shows the chapter estimate and offers the download.

**What Page Sync adds** (`Services/PageSync`, `Screens/PageSync`)

- Live scanner (VisionKit `DataScannerViewController`, stubbed on Mac Catalyst) that captures on its own.
- `BookSections`: chapter-level view of the `BookTextIndex` (resource ranges + TOC titles), and `AudioPositionEstimator`: projects a text position onto the audio chapters and produces candidate times, with a per-book drift calibration.
- `LibraryItemPairing` / `ReadListenPairing`: finds the ebook item for an audiobook when they are separate library items with the same title. This might be useful for Read Along too, which currently requires both formats in one item.
- The sheet, its view model, the intro sheet, and the player menu entry.
- Info.plist: `NSCameraUsageDescription`; the speech description now mentions both features.

**Testing**

Tested on iPhone with iOS 26 against Audiobookshelf 2.x, with a single-file M4B audiobook and its EPUB, downloaded. Also verified the failure paths: no ebook in the library, page not found, audiobook not downloaded.

**Notes for review**

- Follows the View / `@Observable` Model / ViewModel pattern and `swift-format`.
- Strings added to `Localizable.xcstrings` in English only.
- Happy to gate the entry behind a preference if you prefer.

Demo (earlier build, same flow): https://github.com/user-attachments/assets/179c5a34-444e-42f1-b5dc-2d7a3cf61b08


andrespinones:
Rebased on main and reworked on top of the Read Along services: the page is located in the book with `BookTextIndex` + `TranscriptAligner`, and the audio is transcribed with `NarrationTranscriber`. Our own text index, matcher and transcriber are gone; the diff is now ~2,000 lines, all additive. Description updated.

PR #370, opened 2026-09-07
Fix read along getting stuck listening for your place

MalcomnM:
## Summary

Read Along could get stuck on **"Listening for your place"** when the speech analyzer stopped unexpectedly.

The feed loop would keep running against an analyzer that had already finished, so the coordinator never saw a failure and couldn't recover. This also led to repeated `SFSpeechErrorDomain Code=16` / `Maximum number of simultaneous requests reached` errors.

This PR detects that condition earlier and retries the transcription session instead of hanging.

## Root cause

`NarrationTranscriber.feed()` had no way to know the analyzer had stopped.

When the analyzer failed, the feed loop continued sending audio and calling `finalize(through:)`, producing repeated:

`SpeechAnalyzer: finalize(through:) call on finished analyzer; ignoring`

The actual error only surfaced through the results stream, which wasn't being awaited until after the feed loop finished. From the coordinator's point of view, nothing had failed, so the popup stayed open and silence recovery could eventually start another session that hit the same error.

### `NarrationTranscriber`

* Runs the audio feed and results stream together so an unexpected analyzer stop fails immediately.
* Treats `analyzerStopped` and Code 16 as transient errors that can be retried.
* Adds logging when the results stream ends.

### `ReadAlongCoordinator`

* Retries transient failures up to four times with `2s, 4s, 8s, 16s` backoff.
* Keeps the UI in `.locating` and retries from the current playhead.
* Prevents silence recovery from starting a competing session.
* Non-transient errors still fail normally.

## Testing

Tested on a physical device running iOS 26.

Before the change, I could reproduce the stuck popup along with Code 16 and repeated `finalize on finished analyzer` logs.

After the change, Read Along recovers normally, including after seeks and transient analyzer failures.


jeremygrenier:
Thank you for the PR it helped me a lot with debugging the issue. I adjusted my last commit to address it. 🙇‍♂️

PR #372, opened 2026-09-07
Fix main-thread hang on Home from SwiftData progress observations

### LINKED ISSUE #371: Home freezes ~5s and the app is killed on a podcast-only library (~44k main-thread SQLite fetches per sync)
## Beta Check

Tested against `main` @ `0257364` (v1.12) built from source, rather than the TestFlight build —
`0257364` is current with `origin/main` as of filing, so it should be at or ahead of the latest beta.
Happy to re-verify on TestFlight if the builds differ meaningfully.

## Description

On a podcast-only library, scrolling the Home screen blocks the main thread for over 5 seconds and the
app is then killed by KSCrash's deadlock watchdog (`"type": "deadlock"`, SIGABRT, past
`deadlockWatchdogInterval = 5.0`). So it isn't a crash in the usual sense — there's one underlying bug,
a main-thread hang.

The main thread spends 100% of its time in `PersistentModel.observe(where:equals:)`. For each
`ModelContext.didSave`, every live observer walks all the changed `PersistentIdentifier`s and reads a
persisted property off each one. On a SwiftData `@Model` that read **faults the object in**, which runs
a full `NSManagedObjectContext.fetch` with its own `sqlite3_prepare_v3`.

On my account that's **1,708 `MediaProgress` rows x 26 observing Home cards ≈ 44,000 synchronous SQLite
fetches on the main thread, per sync.**

Possibly the same underlying problem as #239 (closed as completed), which reported the same
`Application main thread deadlocked` / SIGABRT signature along with the device overheating — consistent
with a main-thread spin. A comment there reported it again after that issue was closed. This report
identifies a specific root cause for that signature.

## Steps to Reproduce

1. Connect to a podcast-only Audiobookshelf server with a large listening history (~1,700 progress records).
2. Open the app on the Home tab.
3. Scroll the Home view.

## Expected Behavior

Home scrolls smoothly.

## Actual Behavior

UI freezes for ~5 seconds, then the app is killed. Reproducible on both device and simulator.

## Device Information

- **Platform:** iPhone
- **Device Model:** iPhone 17 Pro (physical), and iPhone 17 Pro simulator
- **OS Version:** iOS 26.6.1 (device), iOS 26.5 (simulator)
- **AudioBooth Version:** 1.12 (`main` @ `0257364`, source build)
- **Audiobookshelf Server Version:** 2.36.0
- **Library:** single library, `mediaType: podcast` — no audiobooks at all

### Cause

`Models/Sources/Models/PersistentModel+Observations.swift:47` reads a persisted property to test each
changed object:

```swift
for identifier in allChanges {
  guard
    ...
    model[keyPath: keyPath] == value   // line 47
```

That read faults the object in, and the whole loop runs on the MainActor (line 12). 100% of main-thread
samples during the freeze:

```
607  PersistentModel.observe(where:equals:)  PersistentModel+Observations.swift:47
583   MediaProgress.bookID.getter
296    NSManagedObjectContext.fetch -> executeFetchRequest
 60     sqlite3_prepare_v3 -> sqlite3RunParser
```

### Why podcast-only makes it bad

`PodcastCardModel` starts an observer whenever `sortBy == nil` (`PodcastCardModel.swift:71`), which is
how every Home section builds it. `BookCardModel` never observes at all. So on a podcast library all 26
Home cards observe the same table; on an audiobook library almost none would. That may be why this
hasn't shown up more widely.

### Suggested fix

1. **Don't fault just to compare.** Keep the target's `persistentModelID` from the initial fetch and
   compare identifiers in the notification loop, falling back to a fetch only when there are inserts and
   no target is known yet. Turns ~44,000 SQL statements into ~44,000 pointer comparisons.

2. **Cancel the observers.** `ContinueListeningBookCardModel`, `PodcastCardModel` and `BookPlayerModel`
   never cancel `progressObservation`, so every card model ever built leaves a permanent MainActor
   observer and the count keeps growing. `BookDetailsViewModel.swift:45` already does this correctly.

Either one alone should stop the hang. Happy to open a PR if that's useful.

<details>
<summary>Supporting detail</summary>

**CPU during a reproduction**, captured on the simulator (1s samples, killed at the end). Main thread was
100% busy with zero idle in the last three:

```
cpu=0.4%  0.0%  62.3%  2.8%  112.3%  192.9%  161.9%  152.2%  155.1%  162.0%  <- killed
```

**Why all 1,708 rows end up in one notification:** `MediaProgress.update(from:)` assigns `id`,
`duration` and `startedAt` unconditionally (lines 147-153) *before* the `if remoteLastUpdate > lastUpdate`
freshness check on line 164. Assigning a persisted property dirties the object even when the value is
unchanged, so every row `syncFromAPI` touches lands in `NSUpdatedObjectsKey`. Guarding those assignments
would shrink every `didSave`.

**What grows the observer count:** `ContentView.swift:107` builds `HomePage(model: HomePageModel())`
inline in `body`, while every other tab uses a stored `@StateObject`. `@StateObject` keeps only the
first instance, but each discarded one is fully constructed first — and `HomePageModel.init` rebuilds
all 26 card models (`existingModels = [:]` at `HomePageModel.swift:291`), each spawning another
observer.

**Store contents** — only `MediaProgress` has rows, consistent with podcast-only and nothing downloaded:
`ZMEDIAPROGRESS` 1,708; `ZLOCALBOOK`, `ZLOCALEPISODE`, `ZLOCALPODCAST`, `ZPLAYBACKSESSION`, `ZBOOKMARK`,
`ZDOWNLOADREQUEST`, `ZCHAPTER`, `ZTRACK`, `ZAUTHOR`, `ZSERIES` all 0.

**Home sections rendered** (all in `HomeSection.defaultCases`): `continue-listening` 5,
`newest-episodes` 10, `recently-added` 1, `listen-again` 10 = 26 cards.

**Minor, unrelated:** a second thread was also pegged at 100% in
`__generateCulledBacktrace -> dladdr -> findClosestSymbol`, which is KSCrash's zombie monitor
(`monitors = .all`) symbolicating on every `NSObject` dealloc. Not the cause, but it roughly doubles the
cost while this is happening.

</details>

rspeicher:
Fixes #371.

On a podcast-only library the Home screen blocks the main thread long enough that KSCrash's deadlock
watchdog terminates the app. Two independent causes, one commit each — either can be taken on its own.

### Avoid faulting every changed model in SwiftData observations

`PersistentModel.observe(where:equals:)` read a persisted property off every changed object to test it:

```swift
model[keyPath: keyPath] == value
```

On a SwiftData `@Model` that read faults the object in, so each one costs a separate
`NSManagedObjectContext.fetch` (with its own `sqlite3_prepare_v3`), and the whole loop runs on the
MainActor. With ~1,700 `MediaProgress` rows and 26 observing Home cards that's roughly 44,000 synchronous
SQLite fetches per sync.

This resolves the match once and tracks it by `PersistentIdentifier`, so the notification loop compares
identifiers instead of touching the store. A fault knows its identity without being fulfilled, so this
answers the same question with no I/O. It falls back to re-resolving only when a row of that entity is
inserted and nothing is being tracked yet.

### Cancel media progress observations when the owning models are released

`ContinueListeningBookCardModel`, `PodcastCardModel` and `BookPlayerModel` start
`progressObservation = Task { ... }` but never cancel it. Releasing a `Task` handle doesn't cancel the
task, so each one stayed parked on `ModelContext.didSave` for the life of the process, and every rebuild
of the Home sections added more. `BookDetailsViewModel` already cancelled its observations in `deinit`;
this applies the same pattern to the other three.

### Verification

Built and run on the iPhone 17 Pro simulator (iOS 26.5) against the same account from #371 (single
podcast library, 1,708 `MediaProgress` rows).

| | before | after |
|---|---|---|
| CPU after launch | ramps to ~192%, stays ~155% | brief launch spike, then 0.0% at rest |
| Scrolling Home | main thread 100% busy, 0% idle | ~16%, settles immediately |
| Outcome | killed by the watchdog in ~14s | stable |

`PersistentModel.observe` no longer appears anywhere in main-thread samples after launch.

### Not addressed here

Two related things I left alone, since they're behaviour changes rather than fixes to this hang — happy
to add either if you'd like them in scope:

- `MediaProgress.update(from:)` assigns `id`, `duration` and `startedAt` before the
  `remoteLastUpdate > lastUpdate` check, which dirties every row `syncFromAPI` touches and puts all of
  them in each `didSave`.
- `ContentView` builds `HomePage(model: HomePageModel())` inline in `body` while the other tabs use a
  stored `@StateObject`. That's what made the leaked observers compound.

### DIFF
--- a/Models/Sources/Models/PersistentModel+Observations.swift
+++ b/Models/Sources/Models/PersistentModel+Observations.swift
@@ -12,26 +12,17 @@ extension PersistentModel {
       let task = Task { @MainActor in
         let ctx = ModelContextProvider.shared.context
         let descriptor = FetchDescriptor<Self>()
-
-        // Reading a persisted property off a changed model faults it in, costing a separate
-        // store fetch per object. Resolve the match once, then track it by identifier so the
-        // notification loop compares identifiers instead of touching the store.
         var targetID: PersistentIdentifier?
 
-        func locateTarget() -> Self? {
-          do {
-            let items = try ctx.fetch(descriptor)
-            return items.first { $0[keyPath: keyPath] == value }
-          } catch {
-            AppLogger.persistence.error("Failed to fetch \(entityName) for observation: \(error)")
-            return nil
+        do {
+          let items = try ctx.fetch(descriptor)
+          if let model = items.first(where: { $0[keyPath: keyPath] == value }) {
+            targetID = model.persistentModelID
+            nonisolated(unsafe) let model = model
+            continuation.yield(model)
           }
-        }
-
-        if let model = locateTarget() {
-          targetID = model.persistentModelID
-          nonisolated(unsafe) let model = model
-          continuation.yield(model)
+        } catch {
+          AppLogger.persistence.error("Failed to fetch \(entityName) for observation: \(error)")
         }
 
         for await notification in NotificationCenter.default.notifications(
@@ -42,30 +33,35 @@ extension PersistentModel {
             let userInfo = notification.userInfo
           else { continue }
 
-          let inserts = Set((userInfo[NSInsertedObjectsKey] as? [PersistentIdentifier]) ?? [])
-          let updates = Set((userInfo[NSUpdatedObjectsKey] as? [PersistentIdentifier]) ?? [])
-          let deletes = Set((userInfo[NSDeletedObjectsKey] as? [PersistentIdentifier]) ?? [])
+          let inserts = (userInfo[NSInsertedObjectsKey] as? [PersistentIdentifier]) ?? []
+          let updates = (userInfo[NSUpdatedObjectsKey] as? [PersistentIdentifier]) ?? []
+          let deletes = (userInfo[NSDeletedObjectsKey] as? [PersistentIdentifier]) ?? []
 
           if let identifier = targetID, deletes.contains(identifier) {
             targetID = nil
           }
 
           if let identifier = targetID {
-            guard inserts.contains(identifier) || updates.contains(identifier) else { continue }
             guard
+              inserts.contains(identifier) || updates.contains(identifier),
               let matched = modelContext.model(for: identifier) as? Self,
               !matched.isDeleted
             else { continue }
 
             nonisolated(unsafe) let model = matched
             continuation.yield(model)
-          } else if inserts.contains(where: { $0.entityName == entityName }) {
-            // Nothing is being tracked yet, so only a newly inserted row can start
-            // matching. Re-resolve once rather than per changed object.
-            if let model = locateTarget() {
-              targetID = model.persistentModelID
-              nonisolated(unsafe) let model = model
+          } else {
+            for identifier in inserts where identifier.entityName == entityName {
+              guard
+                let matched = modelContext.model(for: identifier) as? Self,
+                !matched.isDeleted,
+                matched[keyPath: keyPath] == value
+              else { continue }
+
+              targetID = identifier
+              nonisolated(unsafe) let model = matched
               continuation.yield(model)
+              break
             }
           }
         }
--- a/Models/Sources/Models/PersistentModel+Observations.swift
+++ b/Models/Sources/Models/PersistentModel+Observations.swift
@@ -13,15 +13,25 @@ extension PersistentModel {
         let ctx = ModelContextProvider.shared.context
         let descriptor = FetchDescriptor<Self>()
 
-        do {
-          let items = try ctx.fetch(descriptor)
-          let model = items.first { $0[keyPath: keyPath] == value }
-          if let model {
-            nonisolated(unsafe) let model = model
-            continuation.yield(model)
+        // Reading a persisted property off a changed model faults it in, costing a separate
+        // store fetch per object. Resolve the match once, then track it by identifier so the
+        // notification loop compares identifiers instead of touching the store.
+        var targetID: PersistentIdentifier?
+
+        func locateTarget() -> Self? {
+          do {
+            let items = try ctx.fetch(descriptor)
+            return items.first { $0[keyPath: keyPath] == value }
+          } catch {
+            AppLogger.persistence.error("Failed to fetch \(entityName) for observation: \(error)")
+            return nil
           }
-        } catch {
-          AppLogger.persistence.error("Failed to fetch \(entityName) for observation: \(error)")
+        }
+
+        if let model = locateTarget() {
+          targetID = model.persistentModelID
+          nonisolated(unsafe) let model = model
+          continuation.yield(model)
         }
 
         for await notification in NotificationCenter.default.notifications(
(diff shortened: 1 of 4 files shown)


Models/Sources/Models/PersistentModel+Observations.swift
```diff
@@ -32,23 +42,31 @@ extension PersistentModel {
             let userInfo = notification.userInfo
           else { continue }
 
-          let inserts = (userInfo[NSInsertedObjectsKey] as? [PersistentIdentifier]) ?? []
-          let updates = (userInfo[NSUpdatedObjectsKey] as? [PersistentIdentifier]) ?? []
-          let deletes = (userInfo[NSDeletedObjectsKey] as? [PersistentIdentifier]) ?? []
+          let inserts = Set((userInfo[NSInsertedObjectsKey] as? [PersistentIdentifier]) ?? [])
+          let updates = Set((userInfo[NSUpdatedObjectsKey] as? [PersistentIdentifier]) ?? [])
+          let deletes = Set((userInfo[NSDeletedObjectsKey] as? [PersistentIdentifier]) ?? [])
```
jeremygrenier: No reason to convert it to a set, each contains is only perform once.

Models/Sources/Models/PersistentModel+Observations.swift
```diff
@@ -13,15 +13,25 @@ extension PersistentModel {
...
-        do {
-          let items = try ctx.fetch(descriptor)
-          let model = items.first { $0[keyPath: keyPath] == value }
-          if let model {
-            nonisolated(unsafe) let model = model
-            continuation.yield(model)
+        // Reading a persisted property off a changed model faults it in, costing a separate
+        // store fetch per object. Resolve the match once, then track it by identifier so the
+        // notification loop compares identifiers instead of touching the store.
+        var targetID: PersistentIdentifier?
```
jeremygrenier, line 15: I understand the intention, and I feel it may indeed be a better approach. My main concern is that I did a lot of testing across different iOS versions and remember encountering some strange behavior. That’s why looping each time felt safer at the time. However, I’ve changed a lot in the app since then, so I’m not sure if that concern still applies.

Models/Sources/Models/PersistentModel+Observations.swift
```diff
@@ -32,23 +42,31 @@ extension PersistentModel {
...
             else { continue }
 
-            nonisolated(unsafe) let value = model
-            continuation.yield(value)
+            nonisolated(unsafe) let model = matched
+            continuation.yield(model)
+          } else if inserts.contains(where: { $0.entityName == entityName }) {
+            // Nothing is being tracked yet, so only a newly inserted row can start
+            // matching. Re-resolve once rather than per changed object.
+            if let model = locateTarget() {
```
jeremygrenier: I think we should avoid using `locateTarget` which is calling `ctx.fetch(descriptor)` as it is a really expensive call.

The following was way cheaper than using `ctx.fetch`.

```swift
  for identifier in inserts where identifier.entityName == entityName {
    guard let model = modelContext.model(for: identifier) as? Self,
          !model.isDeleted,
          model[keyPath: keyPath] == value
    else { continue }
    targetID = identifier
    continuation.yield(model)
    break
  }
```

Models/Sources/Models/PersistentModel+Observations.swift
```diff
@@ -13,15 +13,25 @@ extension PersistentModel {
...
+        // store fetch per object. Resolve the match once, then track it by identifier so the
+        // notification loop compares identifiers instead of touching the store.
+        var targetID: PersistentIdentifier?
+
+        func locateTarget() -> Self? {
+          do {
+            let items = try ctx.fetch(descriptor)
+            return items.first { $0[keyPath: keyPath] == value }
+          } catch {
+            AppLogger.persistence.error("Failed to fetch \(entityName) for observation: \(error)")
```
jeremygrenier: If this catch happens for any reason, the `observe` is most likely dead because if the object already exists, `inserts` will never be true, so `targetID` will never be set correctly.

I think it’s fine, though, because if `try ctx.fetch(descriptor)` fails, something is already quite wrong in the first place.

Models/Sources/Models/PersistentModel+Observations.swift
jeremygrenier: I know this brings the loop back, but at least it will only loop for observers that don’t have a matching object yet, which should greatly reduce the number of observers affected.

jeremygrenier (commented):
Thank you for the contribution. This is such a critical piece of the application.

Can you make sure to test it on different iOS versions, as well as different scenarios, such as logging in for the first time without any `MediaProgress` yet versus opening the app with an already populated `MediaProgress`?

Also please drop the inline comments. They were useful during the review but they don’t need to be part of the codebase.

Models/Sources/Models/PersistentModel+Observations.swift
rspeicher: Switched to exactly that loop. It only runs for observers with no tracked target, so on a populated library it is effectively free.

Models/Sources/Models/PersistentModel+Observations.swift
rspeicher: Agreed. With the insert loop in place, an observer whose initial fetch failed still picks up a later insert, so this is now more resilient than the fetch-based version.

Models/Sources/Models/PersistentModel+Observations.swift
rspeicher: I tested on 26.5, 18.6 and 17.5 with both an empty store and a populated one (details in the summary comment below). On 17.5 `ModelContext.didSave` is never posted at all, so no observer path fires there regardless of the loop body. That is the same before and after this change, and Home still shows progress on 17 via the `onAppear` fetch.

rspeicher:
Addressed in 549d2de: dropped the `Set` conversions, replaced the insert-path fetch with the `model(for:)` loop, and removed the inline comments.

### Verification

Simulators, same account as #371 (podcast-only library, 1,708 `MediaProgress` rows). Each runtime was tested with an already populated store and with the store removed so the app syncs from nothing, which is what a first login looks like.

| Scenario | iOS 26.5 | iOS 18.6 | iOS 17.5 |
|---|---|---|---|
| Populated store, relaunch | stable, 0% CPU at rest | stable, 0% | stable, 0% |
| Empty store, first sync | 1,708 rows synced, stable | same | same |
| Insert path fires (temporary debug log) | 9 of 16 live observers matched | 9 of 16 | never |
| Update path fires on next sync | 9 tracked yields | 9 | never |

One finding worth knowing about: on iOS 17.5 every observer task starts, but `ModelContext.didSave` is never posted during the sync, so the observation layer is inert there. That predates this PR since the original loop waited on the same notification. Home still shows progress on 17 because each card's `onAppear` fetches directly.

The debug logging used for the observer counts was local only and is not in the branch.

jeremygrenier (approved):
Thank you for your contribution and great work!

rspeicher:
@jeremygrenier Thank you for the helpful review and quick turnaround. I'm excited to actually be able to use the app!

Issue #364, opened 2026-08-31
Translation


K1ndWha1e:
Hey! May I add the Russian translation?

jeremygrenier:
@K1ndWha1e for sure. I can add an AI-generated version first, and then you can adjust it as needed if that would make your life easier.

K1ndWha1e:
No thanks. I'll do it myself.

jeremygrenier:
I'm closing the issue as you are working on it. Feel free to open more PR if need be!

PR #377, opened 2026-09-11
*first attempt for russian localization (#364)

K1ndWha1e:
So, this is first PR for Russian localization. More PRs with typos fixes are coming :)

And one more thing:
`variations.plural` (`one` / `few` / `many` / `other`) — these are the counter strings marked `^[%lld book](inflect: true)` and friends. Apple's automatic inflection doesn't produce correct Russian forms, and Russian needs three distinct forms anyway (1 книга / 2 книги / 5 книг), so explicit plural variations are used instead of the inflection markup.

### DIFF
--- a/AudioBooth/AudioBooth/Localizable.xcstrings
+++ b/AudioBooth/AudioBooth/Localizable.xcstrings
@@ -33,6 +33,12 @@
             "value" : ""
           }
         },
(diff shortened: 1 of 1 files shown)


jeremygrenier (approved):
Thank you for the contribution.

PR #383, opened 2026-09-16
Fix context menu sheets in Continue Listening cover flow

octopotato:
Long pressing a book in the Continue Listening cover flow gives you the same menu as the card and list layouts, but the actions that open a sheet did nothing (ex Add to Collection/Playlist), while Play and Download kept working.

| Before | After |
|--------|--------|
| https://github.com/user-attachments/assets/42a112e3-9bc9-49b5-a3a0-9b54b9e4c677 | https://github.com/user-attachments/assets/57ffb527-e28d-4a41-b7d3-5597980d2769 |


jeremygrenier:
@octopotato thank you for the PR 🙇‍♂️  I totally missed it when implementing the cover flow for continue listening.

I went with a slightly different approach [here](https://github.com/AudioBooth/AudioBooth/commit/d0f87ad36923dc78632d5a4752066f8c2ece945e).

octopotato:
nice, I considered that solution but wanted to limit the changes.. I like your change 😄

